PR #4, opened 2020-09-02
Doc fix build

tacaswell:


dmgav:
I guess the only way to see if `doctr` configuration works is to merge the PR.

PR #30, opened 2020-09-08
ENH: pause plan using 'RE.request_pause()' instead of SIGINT

dmgav:
Small improvement in the code: pausing a plan using `RE.request_pause()` instead of generating `SIGINT`. To test the operation the following code may be changes:

Start the server:
```
python -m aiohttp.web -H 0.0.0.0 -P 8080 bluesky_queueserver.server:init_func
```
In a different terminal load a plan (runs for 10 seconds):
```
http POST 0.0.0.0:8080/add_to_queue plan:='{"name":"count", "args":[["det1", "det2"]], "kwargs":{"num":10, "delay":1}}' 
```
Create the environment and start the queue execution (the queue contains one plan):
```
http POST 0.0.0.0:8080/create_environment
http POST 0.0.0.0:8080/process_queue
```
Pause the plan while it is running:
```
http POST 0.0.0.0:8080/re_pause option="immediate"
```
Resume the plan:
```
http POST 0.0.0.0:8080/re_continue option="resume"
```
Check that the last measurement was repeated after the scan is restarted. Now load the plan again (queue can be filled with as many plans as desired). Start the queue execution. Request the deferred pause:
```
http POST 0.0.0.0:8080/re_pause option="deferred"
```
Resume the plan and check that the last measurement in the plan was not repeated after the plan is resumed.
```
http POST 0.0.0.0:8080/close_environment
```

### DIFF
--- a/bluesky_queueserver/worker.py
+++ b/bluesky_queueserver/worker.py
@@ -266,7 +266,7 @@ class RunEngineWorker(Process):
                         raise RuntimeError(f"Option '{option}' is not supported. "
                                            f"Available options: {pausing_options}")
 
-                    defer = (option == "deferred")
+                    defer = {'deferred': True, 'immediate': False}[option]
                     self._RE.request_pause(defer=defer)
                     msg_ack["value"]["status"] = "accepted"
                 except Exception as ex:
--- a/bluesky_queueserver/worker.py
+++ b/bluesky_queueserver/worker.py
@@ -2,8 +2,6 @@ from multiprocessing import Process
 import threading
 import queue
 import time as ttime
-import os
-import signal
 from collections.abc import Iterable
 
 from bluesky import RunEngine
@@ -53,9 +51,6 @@ class RunEngineWorker(Process):
 
         # The thread that receives packets from the pipe 'self._conn'
         self._thread_conn = None
-
-        # Thread used for to send the second sigint after short timeout
-        self._thread_sigint = None
 
         self._db = db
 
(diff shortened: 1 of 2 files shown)


bluesky_queueserver/worker.py
```diff
@@ -262,35 +257,17 @@ def _conn_received(self, msg):
...
-                    #       It seems to work in this prototype, but it needs to be well understood
-                    #       before it can be used in production.
-                    if option == "immediate":
-                        # Run it in a separate thread. So that the function could return immediately.
-                        self._thread_sigint = threading.Thread(target=send_second_sigint,
-                                                               name="RE Worker 2nd SIGINT",
-                                                               daemon=True)
-                        self._thread_sigint.start()
 
+                    defer = (option == "deferred")
```
tacaswell, line 269:
```suggestion
                    defer = {'deferred': True, 'immediate': False}[option]
```

bluesky_queueserver/worker.py
```diff
@@ -262,35 +257,17 @@ def _conn_received(self, msg):
             # Stop the loop in main thread
             logger.info("Pausing Run Engine")
             pausing_options = ("deferred", "immediate")
+            # TODO: the question is whether it is possible or should be allowed to pause a plan in
```
tacaswell, line 260: The state machine on the RE does not allow moving from idle -> paused.

dmgav: I meant the transitional states like 'pausing', 'resuming' etc. The possible issue may be that in the unlikely case the command to pause is issued during any of those states, it will be rejected. So depending on how the rest of the system works, the request has to be either retried or the user has to be informed that the request was rejected. But I think processing of those cases may be left for later when we work on exchange protocols.

PR #34, opened 2020-09-11
Prototyping of the Watchdog process

dmgav:
The first implementation of the Watchdog process that monitors the loop of RE Manager. In the new architecture RE Manager and RE Worker are started by the Watchdog process; RE Worker is started by request from RE Manager. Watchdog holds references to RE Worker and RE Manager. Data exchange between RE Worker and RE Manager is performed only using pipes.

The new command `kill_manager` is implemented in RE Manager and CLI tool (only for testing):
```
qserver -c kill_manager
```
This command blocks the event loop of RE Manager tripping the Watchdog, which restarts RE Manager in 5 seconds. After restart, RE Manager can communicate with the existing RE Worker (if the environment was created). The queue remains unchanged during the restart since it is stored in Redis. Currently restart will not alter operation off the worker when RE is idle, when a plan is running, when the plan is paused. The only known deficiency is that if execution of a plan is finished while manager is unresponsive, the execution of the next plan in the queue will not be started automatically.


bluesky_queueserver/manager.py
```diff
@@ -32,65 +29,107 @@
 """
 
 
-class RunEngineManager:
+class RunEngineManager(Process):
 
-    def __init__(self):
+    def __init__(self, *, conn_watchdog, conn_worker):
```
danielballan, line 34: Should this extend rather than replace the argument accepted by `Process`, or else use composition rather than inheritance?

bluesky_queueserver/manager.py
```diff
@@ -329,6 +402,11 @@ async def _print_db_uids_handler(self, request):
         self._print_db_uids()
         return {"success": True, "msg": ""}
 
+    async def _kill_manager_handler(self, request):
+        # This is expected to block the event loop forever
+        while True:
+            pass
```
danielballan, line 408: Should there be a sleep in this loop to avoiding pinning a CPU?

bluesky_queueserver/manager.py
```diff
@@ -387,10 +474,19 @@ async def zmq_server_comm(self):
...
+    def run(self):
+        """
+        Overrides the `run()` function of the `multiprocessing.Process` class. Called
+        by the `start` method.
+        """
+        logger.info("Starting RE Manager process")
+        try:
+            asyncio.run(self.zmq_server_comm())
+        except Exception as ex:
+            logger.info("Exiting RE Manager with exception %s" % str(ex))
```
danielballan, line 488: The `logger.exception` method includes the whole traceback below the message.

```suggestion
            logger.exception("Exiting RE Manager")
```

bluesky_queueserver/start_manager.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _receive_packet_thread(self):
+        while True:
+            if self._watchdog_to_manager_conn.poll(0.1):
+                try:
+                    msg = self._watchdog_to_manager_conn.recv()
+                    # Messages should be handled in the event loop
+                    self._conn_received(msg)
+                except Exception as ex:
+                    logger.error("Exception occurred while waiting for packet: %s" % str(ex))
```
danielballan, line 50: See note about `logger.exception`

bluesky_queueserver/manager.py
```diff
@@ -162,34 +203,53 @@ def _print_db_uids(self):
...
-            if self._manager_conn.poll(0.1):
+            if self._watchdog_conn.poll(0.1):
                 try:
-                    msg = self._manager_conn.recv()
+                    msg = self._watchdog_conn.recv()
+                    logger.debug(f"Message Watchdog->Manager received: '{msg}'")
+                    # Messages should be handled in the event loop
+                    self._loop.call_soon_threadsafe(self._conn_watchdog_received, msg)
+                except Exception as ex:
+                    logger.error(f"Exception occurred while waiting for packet: {ex}")
```
danielballan, line 215: See note about `logger.exception`

danielballan (commented):
Some comments for consideration. Feel free to ignore.

bluesky_queueserver/manager.py
dmgav: Actually I didn't know about `logger.exception`. I will think about how to use it. Actually there is some work on logging that needs to be done.

bluesky_queueserver/manager.py
dmgav: Yes, sleep will work.

bluesky_queueserver/start_manager.py
```diff
@@ -0,0 +1,163 @@
+from multiprocessing import Pipe
+import threading
+import time as ttime
+import pprint
+
+from .worker import RunEngineWorker
+from .manager import RunEngineManager
+
+import logging
+logging.basicConfig(level=logging.DEBUG)
```
danielballan, line 10: Now I understand why you were seeing overly-verbose log messages from ophyd. This configuration setting will add a handler that displays DEBUG-level messages from _all_ libraries. Library code should not configure handlers like this, at least not at import time. We did this in earlier versions of our libraries and learned to regret it. (Also see the gray "Note" box in the documentation section https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library for some official validation of that opinion.)

This could be added inside the `if __name__ == '__main__':` block (which is more an application than a library) and perhaps be configurable via a verbosity option like `-v`, as we do [in caproto](https://github.com/caproto/caproto/blob/4033c02da7815f7145c4a42581ddf0f76da4bbca/caproto/commandline/get.py#L140-L145).

bluesky_queueserver/worker.py
```diff
@@ -22,12 +18,14 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
-config_bluesky_logging(level='INFO')
-config_ophyd_logging(level='INFO')
+logging.getLogger('bluesky').setLevel('WARNING')
```
danielballan, line 21: This is a useful workaround in scripts or interactive sessions when loggers get to noisy, but modules should not silence loggers at import time. See previous comment for a better approach.

bluesky_queueserver/manager.py
```diff
@@ -3,12 +3,12 @@
...
+import time as ttime
+import pprint
 
-from .worker import RunEngineWorker
-
-from databroker import Broker
+from .worker import DB
 
 import logging
 logging.basicConfig(level=logging.DEBUG)
```
danielballan, line 14: This should be removed.

danielballan (changes requested):
The `master` branch uses `logging.basicConfig(level='DEBUG')` and should not have. That prompted workaround that doubled down on that approach, and it would be better to roll them back.

dmgav:
@danielballan Thank you for the comments. I guess I start to understand how it works now. Actually logging was waiting to be fixed in this project.

bluesky_queueserver/worker.py
dmgav: Controlling verbosity using -v is useful. I will add CLI options in one of the following PRs.

dmgav:
The last commit implements components necessary to restart RE Manager while plan is running in the worker process without disturbing the plan execution.

dmgav:
Moved code for RE Manager and RE Server to two separate directories. Created entry point `start-re-manager` to start RE manager from command line. Updated README.rst

dmgav:
The first 'real' test works with Python 3.7 and 3.8.

mrakitin (approved):
Tested during a call under @dmgav's supervision :)
Discussed a few improvements, which will be implemented in follow-up PRs.

bluesky_queueserver/manager/manager.py
```diff
@@ -0,0 +1,655 @@
...
+    Parameters
+    ----------
+    conn_watchdog: multiprocessing.Connection
+        One end of bidirectional (input/output) for communication to Watchdog process.
+    conn_worker: multiprocessing.Connection
+        One end of bidirectional (input/output) for communication to RE Worker process.
+    args, kwargs
+        `args` and `kwargs` of the `multiprocessing.Process`
+    """
+    def __init__(self, *args, conn_watchdog=None, conn_worker=None, **kwargs):
```
tacaswell, line 44:
```suggestion
    def __init__(self, *args, conn_watchdog, conn_worker, **kwargs):
```

This will make them required kwargs (py36 feature iirc).

bluesky_queueserver/manager/manager.py
```diff
@@ -0,0 +1,655 @@
...
+
+    async def _run_task(self):
+        """
+        Upload the plan to the worker process for execution.
+        Plan in the queue is represented as a dictionary with the keys "name" (plan name),
+        "args" (list of args), "kwargs" (list of kwargs). Only the plan name is mandatory.
+        Names of plans and devices are strings.
+        """
+        n_pending_plans = await self._r_pool.llen('plan_queue')
+        logger.info("Starting a new plan: %d plans are left in the queue" % n_pending_plans)
```
tacaswell, line 215:
```suggestion
        logger.info("Starting a new plan: %d plans are left in the queue", n_pending_plans)
```

This defers doing the templating until the loggers need it.

tacaswell:
I'll continue my review as a PR ;)

bluesky_queueserver/manager/manager.py
dmgav: I see, I will modify the logging commands in the next PR.

PR #35, opened 2020-09-16
Use `json-rpc` package for communication between RE Manager and Watchdog

dmgav:
Changes in this PR:

- Switched to `json-rpc` for communication between RE manager and Watchdog.

- Implemented communication timeout and simplistic handling of the timeout exception. 

- Added test case that involves restart of RE Manager while the plan is running.

- Implemented changes in code style based on requests for the previous PR (e.g. proper formatting of logging messages for better efficiency).

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -41,12 +41,14 @@ class RunEngineManager(Process):
     args, kwargs
         `args` and `kwargs` of the `multiprocessing.Process`
     """
-    def __init__(self, *args, conn_watchdog=None, conn_worker=None, **kwargs):
+    def __init__(self, *args, conn_watchdog, conn_worker, **kwargs):
 
-        if conn_watchdog is None:
-            raise RuntimeError("Parameter 'conn_watchdog' is not specified or None.")
-        if conn_worker is None:
-            raise RuntimeError("Parameter 'conn_worker' is not specified or None.")
+        if not conn_watchdog:
+            raise RuntimeError("Value of the parameter 'conn_watchdog' is invalid: %s.",
+                               str(conn_watchdog))
+        if not conn_worker:
+            raise RuntimeError("Value of the parameter 'conn_worker' is invalid: %s.",
+                               str(conn_worker))
 
         super().__init__(*args, **kwargs)
         self._re_worker = None
@@ -75,8 +77,10 @@ class RunEngineManager(Process):
 
         self._event_worker_created = None
         self._event_worker_closed = None
-        self._fut_is_worker_alive = None
         self._fut_worker_status = None
+
+        self._fut_watchdog_comm = None  # Future for waiting for messages from watchdog
+        self._event_watchdog_comm = None  # Event which is set when message is expected from watchdog
 
     def _start_conn_threads(self):
         self._thread_conn_watchdog = threading.Thread(target=self._receive_packet_watchdog_thread,
@@ -93,10 +97,9 @@ class RunEngineManager(Process):
         Heartbeat generator for Watchdog (indicates that the loop is running)
         """
         t_period = 0.5
-        msg = {"type": "notification", "value": "alive"}
         while True:
             await asyncio.sleep(t_period)
-            self._watchdog_conn.send(msg)
+            await self._watchdog_send_heartbeat()
 
     # ======================================================================
     #          Communication with Redis
@@ -144,12 +147,15 @@ class RunEngineManager(Process):
         """
         self._event_worker_created = asyncio.Event()
 
-        msg = {"type": "command", "value": "start_re_worker"}
-        self._watchdog_conn.send(msg)
-
-        logger.debug("Waiting for RE worker to start ...")
-        await self._event_worker_created.wait()
-        logger.debug("Worker started successfully.")
+        try:
+            success = await self._watchdog_start_re_worker(timeout=0.5)
+            if not success:
+                raise RuntimeError("Failed to create Worker process")
+            logger.debug("Waiting for RE worker to start ...")
+            await self._event_worker_created.wait()
+            logger.debug("Worker started successfully.")
+        except Exception as ex:
+            logger.exception("Failed to start_Worker: %s", str(ex))
 
     async def _started_re_worker(self):
         # Report from RE Worker received: environment was created successfully.
@@ -160,6 +166,7 @@ class RunEngineManager(Process):
         Closes Run Engine execution environment in orderly way. Running plan needs
         to be stopped before the environment can be closed.
         """
+        success = True
         if self._environment_exists:
             self._event_worker_closed = asyncio.Event()
 
@@ -170,13 +177,16 @@ class RunEngineManager(Process):
             await self._event_worker_closed.wait()
             logger.debug("RE Worker to closed")
 
-            msg = {"type": "command", "value": "join_re_worker"}
-            self._watchdog_conn.send(msg)
-
+            if not await self._watchdog_join_re_worker(timeout=0.5):
+                success = False
+                # TODO: this error should probably be handled differently than this,
+                #   since it may indicate that the worker process is stalled.
+                logger.error("Failed to properly join the worker process. "
+                             "The process may not be properly closed.")
             self._environment_exists = False
-            return True
         else:
-            return False
+            success = False
+        return success
 
     async def _stopped_re_worker(self):
         # Report from RE Worker received: environment was closed successfully.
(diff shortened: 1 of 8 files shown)


dmgav:
It seems that the PR is ready.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -34,103 +36,71 @@ def _create_conn_pipes(self):
...
-    def _join_re_worker(self):
+        try:
+            self._re_worker = RunEngineWorker(conn=self._manager_conn, name="RE Worker Process")
+            self._re_worker.start()
+            success, err_msg = True, ""
+        except Exception as ex:
+            success, err_msg = False, str(ex)
+        return {"success": success, "err_msg": err_msg}
+
+    def _join_re_worker_handler(self, *, timeout=0.5):
```
MikeHart85, line 56: Just a question for my own benefit, rather than a comment:

This semantic to force keyword-only arguments is used in many places. What is the motivation for restricting to keyword invocations only? Is it just to make sure the calls will still work, even if the function footprint is changed?

MikeHart85 (approved):
Looks good to me. Just a question below.

bluesky_queueserver/manager/manager.py
```diff
@@ -609,35 +612,157 @@ async def zmq_server_comm(self):
...
+                else:
+                    response = None
+                return response
+
+            finally:
+                self._event_watchdog_comm.clear()  # Clear before the exit as well
+
+    async def _watchdog_response(self, response):
+        """
+        Set the fututure with the results
```
mrakitin, line 693:
```suggestion
        Set the future with the results
```

bluesky_queueserver/manager/manager.py
```diff
@@ -609,35 +612,157 @@ async def zmq_server_comm(self):
...
+    async def _watchdog_kill_re_worker(self):
+        """
+        Request Watchdog to kill RE Worker process (justified only if the process is not responsive).
+        """
+        try:
+            response = await self._watchdog_send("kill_re_worker")
+            success = response["result"]["success"]
+        except asyncio.TimeoutError:
+            success = False
+        return success
```
mrakitin, line 738: The 3 functions above are almost identical except the arguments passed to `self._watchdog_send(...)`. Maybe it's worth to generalize?

bluesky_queueserver/tests/test_examples.py
```diff
@@ -1,3 +0,0 @@
-def test_one_plus_one_is_two():
-    "Check that one and one are indeed two."
-    assert 1 + 1 == 2
```
mrakitin, line 3: Should this file be removed completely?

requirements.txt
```diff
@@ -6,3 +6,4 @@ databroker
 ophyd
 pyzmq
 requests
+json-rpc
```
mrakitin, line 9: Can we also add `matplotlib` and `pyqt5` here as well? That way it will be installed without installing requirements from `requirements-dev.txt` as these packages are run dependencies (good for future packaging/deployments).

mrakitin (commented):
Looks good to me. I have a few minor suggestions (comments corrections) and a question about generalizing similar async functions.

bluesky_queueserver/manager/start_manager.py
dmgav: JSON RPC allows for passing arguments as lists, but I think it is safer to pass them as dictionaries. This way the order of the arguments does not matter. `json-rpc` will not accept requests with arguments as lists if handler has keyword-only arguments. But in general the code is more readable if parameter names are specified in function calls, especially in case when libraries are under development. It is easier to relax the restrictions later.

bluesky_queueserver/manager/manager.py
dmgav: I think we should keep it for now. This is temporary solution. Those functions may diverge in the next iteration, because timeout handling could be (or should be) different for those cases. If not, then I will generalize it later.

bluesky_queueserver/tests/test_examples.py
dmgav: Yes. I will delete it as part of the next PR. I thought about leaving the file for future tests without looking at the name.

requirements.txt
dmgav: I am not sure about including `matplotlib` and `pyqt5` into `requirements.txt`. It will try to install `pyqt5` from PyPI every time one tries to install the package from source. We typically have `pyqt` installed from `nsls2forge`. I am thinking of adding `scikit-image` to the requirements.

dmgav: Actually `bluesky-queueserver` doesn't use `PyQt` or `Matplotlib` (at least not yet). Those should be requirements of Bluesky or they should be installed with Bluesky.

bluesky_queueserver/manager/start_manager.py
tacaswell: It protects you against adding / removing arguments and it it (arguably) makes the calls clearer (as the keyword name is in line).  I think this is as strength for internal code where it is mostly read and only rarely typed.

bluesky_queueserver/tests/test_examples.py
tacaswell: It is the cookie-cutter dummy test, seems ok to nuke to me.

requirements.txt
tacaswell: These should definitly not depend on mpl and pyqt

mrakitin: OK, I just noticed it was missing them during my run a few PRs ago. I'll check later if it's still needed for anything in a fresh environment.

bluesky_queueserver/manager/manager.py
mrakitin: Sounds good.

requirements.txt
mrakitin: It's still hitting me:
```py
(qs2) ✘-INT ~/src/mrakitin/bluesky-queueserver [master|✔] 
10:55 $ start-re-manager
Traceback (most recent call last):
  File "/home/mrakitin/miniconda3/envs/qs2/bin/start-re-manager", line 33, in <module>
    sys.exit(load_entry_point('bluesky-queueserver', 'console_scripts', 'start-re-manager')())
  File "/home/mrakitin/miniconda3/envs/qs2/bin/start-re-manager", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/home/mrakitin/miniconda3/envs/qs2/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 105, in load
    module = import_module(match.group('module'))
  File "/home/mrakitin/miniconda3/envs/qs2/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/mrakitin/src/mrakitin/bluesky-queueserver/bluesky_queueserver/manager/start_manager.py", line 6, in <module>
    from .worker import RunEngineWorker
  File "/home/mrakitin/src/mrakitin/bluesky-queueserver/bluesky_queueserver/manager/worker.py", line 16, in <module>
    from bluesky.callbacks.best_effort import BestEffortCallback
  File "/home/mrakitin/miniconda3/envs/qs2/lib/python3.7/site-packages/bluesky/callbacks/best_effort.py", line 11, in <module>
    import matplotlib.pyplot as plt
ModuleNotFoundError: No module named 'matplotlib'
```

dmgav: MPL is needed to run bluesky-queueserver with Bluesky, so it needs to be installed with Bluesky., but Matplotlib is not the dependency for bluesky-queueserver itself. And actually MPL is not listed as a dependency of Bluesky. The tutorial recommends the following command to install Bluesky and its dependencies:
```
conda install -c nsls2forge bluesky ophyd databroker matplotlib pyqt=5 ipython
```

mrakitin: I agree. I wonder why bluesky does not depend on matplotlib then? @tacaswell, what was the reasoning? Maybe we can do it through extra requirements there?

dmgav: I think that the reason is that Bluesky may be used without pyqt and Matplotlib. It depends on callbacks used for data presentation.

mrakitin: I agree. Matplotlib should come as a dependency of bluesky (`BestEffortCallback` and friends).

We tested it yesterday, and for the queueserver it was enough to install `matplotlib` alone, without `PyQt5`, and the server was functioning properly. Surprisingly, PyQt5 v5.15 did not play nice with Matplotlib. @tacaswell, have you encountered this issue?

PR #38, opened 2020-09-17
Testing for Watchdog module

dmgav:
Some more tests. Covered functionality of Watchdog module. Changed `WatchdogProcess.__init__ ()` so that it now accepts classes for RE Manager and RE Worker as parameters. It makes the design modular (any custom classes can be passed as parameters). This also makes testing much easier, since classes that emulate RE Manager and RE Worker APIs and run in threads may be used instead of the real ones that run as separate processes.  

The PR should be safe to merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ install:
   - pip install .
   # Install extra requirements for running tests and building docs.
   - pip install -r requirements-dev.txt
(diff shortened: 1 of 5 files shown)


mrakitin (approved):
I like the tests and the way `format_jsonrpc_msg` is used throughout.

PR #39, opened 2020-09-18
Loading of simulated beamline profiles.

dmgav:
Simulated beamline profiles are loaded from the directory specified with command-line option `-p` (use `-h`) to see all available options. If the option is not specified, then the 'default' profile collection is loaded. The 'default' profile collection loads all standard simulated Ophyd devices and standard Bluesky plans and intended to be used for testing and evaluation. Parsing of plans is performed from the namespace that contains the loaded devices and plans. The parser is also rewritten for better flexibility. 

Parsing is still performed before the plan is executed. In case a plan can not be parsed, the plan is substituted for a function that will raise exception in the same thread that is supposed to execute the plan. The error message is used to generate the report sent to the manager. In case of error, RE manager stops execution of the queue and leaves the plan which caused a problem at the beginning of the queue. There is no option to remove a plan from the beginning of the queue right now, so the only option right now is to clear the queue completely and load new plans to the queue.

### DIFF
--- a/bluesky_queueserver/manager/start_manager.py
+++ b/bluesky_queueserver/manager/start_manager.py
@@ -2,10 +2,12 @@ import argparse
 from multiprocessing import Pipe
 import threading
 import time as ttime
+import os
 
 from .worker import RunEngineWorker
 from .manager import RunEngineManager
 from .comms import PipeJsonRpcReceive
+from .profile_ops import get_default_profile_collection_dir
 
 from .. import __version__
 
@@ -167,10 +169,11 @@ def start_manager():
         epilog=f"blueksy-queueserver version {__version__}",
     )
     parser.add_argument("--kafka_topic", type=str, help="The kafka topic to publish to.", )
-    parser.add_argument(
-        "--kafka_server", type=str, help="bootstrap server to connect to.",
-        default="127.0.0.1:9092"
-    )
+    parser.add_argument("--kafka_server", type=str, help="Bootstrap server to connect to.",
+                        default="127.0.0.1:9092")
+    parser.add_argument("--profile_collection", "-p", dest="profile_collection_path",
+                        type=str, help="Path to directory that contains profile collection.", )
+
     logging.basicConfig(level=logging.WARNING)
     logging.getLogger('bluesky_queueserver').setLevel("DEBUG")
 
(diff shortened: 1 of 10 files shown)


tacaswell:
I have a slight concern about "parse" here as I think it gives the wrong idea about what we are doing, but :+1: otherwise.

tacaswell (approved):
Slight concern over the use of parse, but @dmgav can self-merge if he is not worried.

dmgav:
@tacaswell Thanks. I will think about a different name. It can always be renamed later.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -181,6 +184,22 @@ def start_manager():
         config['kafka']['topic'] = args.kafka_topic
         config['kafka']['bootstrap'] = args.kafka_server
 
+    if args.profile_collection_path:
+        pc_path = args.profile_collection_path
+        pc_path = os.path.abspath(os.path.expanduser(pc_path))
+        if not os.path.exists(pc_path):
```
mrakitin, line 190: I'd better use `os.path.isdir` here.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -181,6 +184,22 @@ def start_manager():
...
+            logger.error("Profile collection directory '%s' does not exist", pc_path)
+            return 1
+        if not os.path.isdir(pc_path):
+            logger.error("Path to profile collection '%s' is not a directory", pc_path)
+            return 1
```
mrakitin, line 195: I see these 2 checks, I think it's fine to keep them.

mrakitin (commented):
Reviewing it after it was merged :)
Looks good. Regarding the name, would something like `objectize_plan` have more sense here?

bluesky_queueserver/manager/start_manager.py
dmgav: There are two checks because there are two separate error messages.

bluesky_queueserver/manager/start_manager.py
dmgav: I guess I should put two check here as well, because this is where the consistency of command-line parameters should be checked (if possible).

dmgav:
I haven't see a word `objectize` to be used as a technical term. It is a strange word. Let's see. It's not urgent.

bluesky_queueserver/manager/start_manager.py
dmgav: I see, it's the same place. I also do the same checks in the function that loads the profiles, but I think it should be checked in both places.

MikeHart85:
How about `resolve_plan` or `resolve_plan_names`?

Like resolving a symlink or pointer.

mrakitin:
Ah, it's in the same sense as `pythonize`, also not a technical format term, but is used throughout. I like the `resolve` option suggested by Mike too.

PR #42, opened 2020-09-22
Add http server tests

mrakitin:
This PR covers all currently existing aiohttp routes/endpoints.

### DIFF
--- a/bluesky_queueserver/server/tests/test_http_server.py
+++ b/bluesky_queueserver/server/tests/test_http_server.py
@@ -1,16 +1,18 @@
-import sys
-import psutil
-import pytest
 import subprocess
+import sys
+
+import pytest
 import requests
-from bluesky_queueserver.manager.tests.test_general import re_manager
+from xprocess import ProcessStarter
+
 import bluesky_queueserver.server.server as bqss
+from bluesky_queueserver.manager.tests.test_general import \
+    re_manager  # noqa F401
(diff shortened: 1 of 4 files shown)


bluesky_queueserver/server/tests/test_http_server.py
```diff
@@ -0,0 +1,185 @@
...
+                     "{'name':'count', 'args':[['det1', 'det2']], 'kwargs':{'num':10, 'delay':1}}"])
+    subprocess.call(["qserver", "-c", "add_to_queue", "-v",
+                     "{'name':'count', 'args':[['det1', 'det2']]}"])
+    subprocess.call(["qserver", "-c", "add_to_queue", "-v",
+                     "{'name':'count', 'args':[['det1', 'det2']]}"])
+
+    yield
+
+    subprocess.run('qserver -c clear_queue'.split())
+
```
dmgav, line 49:
```suggestion
    try:
        p.wait(5)
    except subprocess.TimeoutExpired:
        # The manager is not responsive, so just kill the process.
        p.kill()
        assert False, "RE Manager failed to stop"
```
I would put this code here. Sometimes the manager is frozen and don't stop. This means there is a bug and it will be indicated as an error in test results. This also kills the process so that other tests could use the manager.

dmgav (approved):
Looks good

mrakitin:
Thanks, @dmgav!

PR #43, opened 2020-09-23
JSON RPC between RE Manager and RE Worker

dmgav:
The code for RE Manager and RE Worker are rewritten to use JSON RPC protocol for communication. The code may need improvement in terms of readability and error handling. Most of the handlers are modified to initiate requested action and return instant response. Additional APIs are needed to poll the state of RE Manager to determine if the requested action is completed.

Tests are modified to work with the new version of RE Manager. Some of the modifications are temporary: those tests should be changed when additional APIs are available.

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -87,15 +87,6 @@ class RunEngineManager(Process):
         self._background_task_status = {"status": "success",
                                         "err_msg": ""}
 
-
-    '''
-    def _start_conn_threads(self):
-        self._thread_conn_worker = threading.Thread(target=self._receive_packet_worker_thread,
-                                                    name="RE QServer Comm2",
-                                                    daemon=True)
-        self._thread_conn_worker.start()
-    '''
-
     async def _heartbeat_generator(self):
         """
         Heartbeat generator for Watchdog (indicates that the loop is running)
@@ -251,7 +242,7 @@ class RunEngineManager(Process):
 
             if not await self._confirm_re_worker_exit():
                 success = False
-                err_msg = f"Failed to confirm closing of RE Worker thread"
+                err_msg = "Failed to confirm closing of RE Worker thread"
 
         return success, err_msg
 
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -224,9 +224,7 @@ class RunEngineManager(Process):
             msg = {"type": "command", "value": "quit"}
             self._worker_conn.send(msg)
 
-            print(f"Preparing to wait for event 'closed'")  ##
             status = await self._fut_worker_closed_waiting
-            print(f"Finished wait")  ##
             success = (status == "accepted")
         else:
             success = False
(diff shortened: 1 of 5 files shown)


bluesky_queueserver/manager/manager.py
```diff
@@ -223,34 +388,47 @@ async def _run_task(self):
...
-                   "value": {"name": plan_name,
-                             "args": args,
-                             "kwargs": kwargs,
-                             "plan_uid": plan_uid
-                             }
-                   }
+            plan_info = {"name": plan_name,
+                         "args": args,
+                         "kwargs": kwargs,
+                         "plan_uid": plan_uid
```
mrakitin, line 394:
```suggestion
                         "plan_uid": plan_uid,
```

bluesky_queueserver/manager/worker.py
```diff
@@ -222,126 +228,196 @@ def plan():
...
+        logger.info("Pausing Run Engine")
+        pausing_options = ("deferred", "immediate")
+        # TODO: the question is whether it is possible or should be allowed to pause a plan in
+        #       any other state than 'running'???
+        err_msg = ""
+        if self._RE._state == 'running':
+            try:
+                if option not in pausing_options:
+                    raise RuntimeError(f"Option '{option}' is not supported. "
+                                       f"Available options: {pausing_options}")
```
mrakitin, line 362: Should this check be done outside of this condition?

bluesky_queueserver/server/tests/test_http_server.py
```diff
@@ -104,45 +106,56 @@ def test_http_server_close_environment_handler(re_manager, aiohttp_server):  # n
...
-    assert resp4 == {'success': True, 'msg': ''}
-    resp4a = _request_to_json('get', '/queue_view')
-    assert len(resp4a['queue']) == 1
+    # resp4 = _request_to_json('post', '/process_queue')
+    # assert resp4 == {'success': True, 'msg': ''}
+    # resp4a = _request_to_json('get', '/queue_view')
+    # assert len(resp4a['queue']) == 1
+
+    # resp5 = _request_to_json('post', '/process_queue')
+    # assert resp5 == {'success': True, 'msg': ''}
```
mrakitin: Ah, I fixed it in one place via https://github.com/bluesky/bluesky-queueserver/pull/42/commits/c9dce7c8dfcce8d27bd1a0b1d4c9b91ca62cbdea, but not here. Thanks for commenting it. I think it can even be removed as it was my wrong understanding of the workflow.

bluesky_queueserver/manager/worker.py
```diff
@@ -404,8 +491,6 @@ def init_namespace():
         bec = BestEffortCallback()
         self._RE.subscribe(bec)
 
-        # db = Broker.named('temp')
```
mrakitin, line 407: Where do we define a databroker instance and subscription now?

mrakitin (approved):
Looks great! I have just a couple of questions.

I pushed a clean-up commit.

bluesky_queueserver/server/tests/test_http_server.py
mrakitin: I cleaned it up via https://github.com/bluesky/bluesky-queueserver/pull/43/commits/164ae805fea0bf4cc3b030fb4ffed639dc81ad8e.

bluesky_queueserver/manager/worker.py
dmgav: I don't think so. First the command should be checked at the manager, so this check should pass if there are no bugs. And RE is also expected to be running to be running when the command is issued if there are no bugs in the manager. It's just another layer of safety that protects RE from being driven into strange condition. In terms of performance, checking if RE is running seems less costly, so we do it first.

bluesky_queueserver/manager/worker.py
dmgav: DB is instantiated in Line 27.

mrakitin (approved):

PR #44, opened 2020-09-23
FastAPI web server

mrakitin:
Summary:
--------------
- Converted all `aiohttp` calls to `fastapi` with corresponding `get`/`post` decorators. That resulted in getting rid of most handler methods from the `WebServer` class, so I use the latter for the ZMQ communications only (the code for those handler methods is commented out for now, but will eventually be removed).
- Used enum validations for `re_pause`/`re_continue` operations.
- Moved the test fixtures for the server into the corresponding `conftest.py` module.
- All tests passed with the existing tests (just with a different web-server fixture, but the same REST API calls).
- Parametrized ZMQ server/port.

**Note**: we have to install `uvicorn` explicitly, or I opted to use `fastapi[all]` which has that dependency.

### DIFF
--- a/bluesky_queueserver/server/server.py
+++ b/bluesky_queueserver/server/server.py
@@ -1,6 +1,7 @@
-from fastapi import FastAPI
+from fastapi import FastAPI, HTTPException
+from enum import Enum
 
-from aiohttp import web
+# from aiohttp import web
 import asyncio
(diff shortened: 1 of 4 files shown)


dmgav:
Looks good. Let me try to rebase my new version and see how it works.

PR #45, opened 2020-09-24
API for more accurate monitoring of RE Manager state

dmgav:
Changes:
- Extended the information returned by 'ping' request. The returned state contains data sufficient to determine if current operation is completed. More convenient APIs may be implemented in the future for waiting for completion of specific operations.
- Changed variable that keeps current state of RE Manager process from string to `super_state_machine.machines.StateMachine` (same as in Bluesky). Currently it is used as enhanced enum (ensures that assigned values are from a fixed set and reduces the chance of an error), but may eventually be used as a state machine when the rest of the code is ready.
- Modified the CLI tests so they are working without artificial delays. The delays are replaced with polling RE Manager state.
- Renamed command/value to method/params in ZMQ-related code to make it consistent with 'json-rpc' naming. Changed option `-v` to `-p (--parameters)` in CLI tool. `Parameters` seem to be better name than `value`. Also `-v` is often used to control verbosity of the output. Option `-c (--command)` was kept unchanged, since it is more suitable for user interface than 'method'. Updated `README.rst` to reflect the changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,9 +163,9 @@ Run 'qserver' in the monitoring mode (send 'ping' request to RE Manager every se
 
 Add a new plan to the queue::
 
-  qserver -c add_to_queue -v '{"name":"count", "args":[["det1", "det2"]]}'
-  qserver -c add_to_queue -v '{"name":"scan", "args":[["det1", "det2"], "motor", -1, 1, 10]}'
-  qserver -c add_to_queue -v '{"name":"count", "args":[["det1", "det2"]], "kwargs":{"num":10, "delay":1}}'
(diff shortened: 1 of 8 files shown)


mrakitin (approved):
Looks good to me. Probably the next step would be to update the http/web server tests to eliminate the sleeps. We can discuss it tomorrow.

dmgav:
I renamed message parameters for ZMQ to make it consistent with `json-rpc`. All tests are passing.

mrakitin (commented):
I like the last massive update from values to parameters.

PR #47, opened 2020-09-25
update webserver run command in readme to suit new fastapi implementation

developingAlex:


dmgav:
@developingAlex Thank you. We are trying to keep documentation (at least `README`) up to date, but we missed this change.

dmgav (approved):

mrakitin:
Thank you! In tests I was using localhost:8000 (default fastapi values). Maybe we can change that too?

tacaswell:
Given that we have a bunch of services we are standing up it is better if we keep them all on their own ports and never use the default port.

PR #49, opened 2020-09-25
Expand installation instructions and requirements

cryos:
The server needs redis installed to run, and matplotlib is required for
the best effort callback currently used to output debug information.
There is a msgpack minimum version of 1.0 for bluesky kafka that the
version specifications fixes in default installs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,7 @@ and then kill the processes::
 
 Installation of QueueServer from source::
 
+  sudo apt install redis
   pip install -e .
 
 This also sets up an entry point for the 'qserver' CLI tool.
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,9 @@ bluesky-kafka
 databroker
 fastapi
 json-rpc
-msgpack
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -38,6 +38,7 @@ and then kill the processes::
 
 Installation of QueueServer from source::
 
+  sudo apt install redis
```
mrakitin, line 41: It's for debian-like system. Should we add more instructions for `yum`/`dnf` installation steps? (I assume no `zypper`, but we may need to also cover Arch Linux).

cryos: As an Arch Linux user I think having the package name is enough and you can translate for your distro. I guess it could be more generic - use your package manager to install redis?

mrakitin: I think so, but maybe leaving the example for Ubuntu as well.

dmgav: Looking into the future, we are designing a package potentially useful by non-NSLS-II Bluesky users. So having installation instructions for as many systems as possible may be useful. At least this could be a reason to keep already existing instructions, just group them properly.

PR #50, opened 2020-09-25
STY: apply black to whole codebase

tacaswell:


mrakitin (approved):
Do we know how to check it in CI that `black` may want change something?

untzag:
if you like black, consider using it with pre-commit

https://github.com/psf/black#version-control-integration

https://github.com/bluesky/yaqc-bluesky/blob/master/.pre-commit-config.yaml

ksunden:
@mrakitin you can run `black --check` (which returns non-zero exit status if something would be reformatted)

Additionally, I tend to add a `pre-commit` for black, which just reformats when you go to commit... requires people to actually use it, but if at least the major contributors are, it keeps it formatted whenever you touch those files.

PR #51, opened 2020-09-26
Enh more profile work

tacaswell:
This includes the style change from #50 .

The primary change here is that it makes the list of devices / plans richer by returning a dict of meta-data which I think will be important for the UI.

This may be of interest to @cryos .


dmgav:
Why did you have to apply Black?

mrakitin:
I guess the black commit is left from #50. Rebasing on main should eliminate it, I suppose.

tacaswell:
Thanks for fixing the CI issue and sorry for causing havoc with black.

PR #54, opened 2020-09-29
Improved implementation of operations with a plan queue

dmgav:
Contents of this PR:

- Another rewrite of class `PlanQueueOperations` with primary purpose of improving the implementation of queue operations.

- All public methods are protected by `asyncio.Lock()`.

- There are no new server (manager) APIs implemented in this PR. RE Server/Manager are operating exactly as before. The next PR will be dedicated to improvement of APIs.

- Implemented 'history' of executed plans. There are no API to access history yet, but there will be API for reading and clearing the history in the next PR.

- Full and detailed unit test coverage of `PlanQueueOperations` class.


bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -85,62 +298,361 @@ async def get_plan_queue(self):
...
+            await self._add_plan_to_history(plan)
+        else:
+            plan = {}
+        return plan
+
+    async def set_processed_plan_as_stopped(self, exit_status):
+        """
+        Pushes currently executed plan to the beginning of the queue and adds
+        it to history with additional sets `exit_status` key.
+        UID is reamains in `self._uid_set`.
```
mrakitin, line 644:
```suggestion
        UID is remains in `self._uid_set`.
```

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
```diff
@@ -0,0 +1,585 @@
...
+
+@pytest.fixture
+def pq():
+    pq = PlanQueueOperations()
+    asyncio.run(pq.start())
+    # Clear any pool entries
+    asyncio.run(pq.delete_pool_entries())
+    yield pq
+    # Don't leave any test entries in the pool
+    asyncio.run(pq.delete_pool_entries())
```
mrakitin, line 19: Should we move this fixture to `conftest.py`? Looks like it should be the place to store the fixtures.

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
```diff
@@ -0,0 +1,585 @@
...
+
+        await pq._set_running_plan_info(some_plan)
+        await pq.delete_pool_entries()
+        assert await pq.get_running_plan_info() == {}
+        assert await pq.is_plan_running() is False
+
+    asyncio.run(testing())
+
+
+# fmt: off
```
mrakitin, line 53: Is it for `black`?

mrakitin (approved):
Looks good! I don't see any major issues with this new rewrite. I have a few minor comments/suggestions to consider.

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
dmgav: Yes. I think that `@pytest.mark.parametrized` dictionaries should be protected from autoformatting, because they grow large. Also they are manually formatted for readability and convenience of editing. Actually this was one of the issues mentioned by @danielballan when we discussed whether `black` should be used for our projects.

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
dmgav: I guess that if it is used only for tests in this file, it should stay in this file. I don't see it being used anywhere else at this point. But it can always be moved if needed. I think the main idea is to avoid test files importing each other.

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -85,62 +298,361 @@ async def get_plan_queue(self):
...
         -------
-        dict or None
-            The first plan in the queue represented as a dictionary or None if the queue is empty.
+        dict
+            Dictionary of plan parameters. Returns `{}` if no plan is found at the index.
+
+        Raises
+        ------
+        TypeError
+            Incorrect text for
```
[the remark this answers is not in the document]
dmgav: Thanks. I corrected it in the code for the next PR.

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
mrakitin: Sounds good, I like this workaround.

bluesky_queueserver/manager/manager.py
```diff
@@ -32,16 +31,12 @@
...
-        IDLE = "idle"
-        CREATING_ENVIRONMENT = "creating_environment"
-        EXECUTING_QUEUE = "executing_queue"
-        CLOSING_ENVIRONMENT = "closing_environment"
+class MState(enum.Enum):
+    INITIALIZING = "initializing"
+    IDLE = "idle"
+    CREATING_ENVIRONMENT = "creating_environment"
+    EXECUTING_QUEUE = "executing_queue"
+    CLOSING_ENVIRONMENT = "closing_environment"
```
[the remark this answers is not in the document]
dmgav: I am not sure. Actually upper-cased attributes are clearly visible in the code, so it's an advantage. Also the examples for `Enum` are using capitalized letters.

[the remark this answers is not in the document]
mrakitin: Indeed, Enum's docs have a lot of capitalized examples in https://docs.python.org/3/library/enum.html. I agree it's more readable in the code, let's stick with it for now.

mrakitin:
I didn't go through all single back-ticks -> double back-ticks changes, but it can be changed later in a follow up PR, so it's not a blocker here. Thanks for accepting the suggestions.

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -1,47 +1,233 @@
...
+        return plan
+
+    # --------------------------------------------------------------------------
+    #                          Operations with UID set
+
+    def _uid_set_clear(self):
+        """
+        Clear `self._uid_set`.
+        """
+        self._uid_set = set()
```
danielballan, line 170: Why not `self._uid_set.clear()`? This avoids the class of bugs where some other part of the program might be holding a reference to a stale versino of `_uid_set`. Experience with this led us to clear rather than recreate all the caches in `RunEngine` and `Bundler`.

Also, I am not convinced that these custom private methods add value over simply calling the methods on the private set itself.

dmgav: I guess you are right. I will change it in the next PR.

dmgav: Actually the idea behind the private methods was that they provide some level of abstraction. If the UIDs are to be stored in some different way (e.g. in Redis), then only those private methods need to be rewritten, not the code using them. So I am not sure if I wouldn't need to put those methods back if I remove them. So for now I will put TODO next to those methods and if we don't need them they could be removed in the future. 
Creating an instance of the set instead of clearing it is not right, so I will change it.

PR #55, opened 2020-09-29
Expansion of RE Manager API

dmgav:
Changes to RE Manager API:

- renamed `queue_view` to `get_queue`. It now returns a dictionary with two keys: `queue` - list of plans in the queue (represented as dictionaries, first element in the list is the first plan to run), `running_plan` - dictionary that represents currently running plan (running plan is not part of the queue, `{}` is returned if no plan is currently executed).

- `clear_queue` entry point to HTTP server (the operation was available via CLI client). The test is also added.

- `get_history` and `clear_history` API (both CLI and HTTP). The tests are added.

### DIFF
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -335,7 +335,7 @@ class PlanQueueOperations:
         Raises
         ------
         TypeError
-            Incorrect text for
+            Incorrect value of `pos` (most likely a string different from `front` or `black`
         """
         async with self._lock:
             return await self._get_plan(pos)
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -101,7 +101,7 @@ class PlanQueueOperations:
 
     async def _delete_pool_entries(self):
         """
-        See ``self.delete_pool_entries`` method.
+        See ``self.delete_pool_entries()`` method.
         """
         await self._r_pool.delete(self._name_running_plan)
         await self._r_pool.delete(self._name_plan_queue)
@@ -212,7 +212,7 @@ class PlanQueueOperations:
 
     async def _is_plan_running(self):
         """
-        See ``self.is_plan_running`` method.
+        See ``self.is_plan_running()`` method.
         """
         return bool(await self._get_running_plan_info())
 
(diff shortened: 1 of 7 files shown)


bluesky_queueserver/server/server.py
```diff
@@ -110,20 +110,20 @@ class REResumeOptions(str, Enum):
 
 
 @app.get("/")
-async def _hello_handler():
+async def _ping_handler():
```
mrakitin, line 113: Maybe we don't need these handler functions to be private, as they are going to be used by FastAPI (only).

bluesky_queueserver/server/server.py
```diff
@@ -151,7 +160,25 @@ async def _pop_from_queue_handler():
...
     """
-    msg = await re_server._send_command(command="pop_from_queue")
+    msg = await re_server.send_command(command="pop_from_queue")
+    return msg
+
+
+@app.get("/get_history")
+async def _get_history_handler():
+    """
+    Returns the plan history (list of dict).
```
mrakitin, line 170:
```suggestion
    Returns the plan history (list of dicts).
```
... or "list of dictionaries".

mrakitin (approved):
Nice! I like the added history API.

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -341,7 +341,7 @@ async def get_plan(self, pos):
         Raises
         ------
         TypeError
-            Incorrect value of ``pos`` (most likely a string different from ``front`` or ``black``
+            Incorrect value of ``pos`` (most likely a string different from ``front`` or ``black``)
```
mrakitin, line 344:
```suggestion
            Incorrect value of ``pos`` (most likely a string different from ``front`` or ``back``)
```

bluesky_queueserver/server/server.py
dmgav: Removed the underscores.

mrakitin: Thanks!

bluesky_queueserver/manager/plan_queue_ops.py
mrakitin: I've self-committed. Once the CI passes, should be good to go.

mrakitin (approved):

PR #56, opened 2020-09-30
WIP: dummy get_ipython()

tacaswell:
Currently fails inside of nslsii.configure_base which directly imports
get_ipython

### DIFF
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -49,10 +49,33 @@ def load_profile_collection(path):
     file_list.sort()  # Sort in alphabetical order
 
     # Load the files into the namespace 'nspace'.
-    nspace = None
+    nspace = {}
+
+    class IPDummy:
+        def __init__(self, user_ns):
+            self.user_ns = user_ns
+
+        def __call__(self):
+            return self
+
+        def register_magics(self, *args, **kwargs):
+            ...
+
+        def magics(
+
+    class DummyRE:
+        preprocessors = []
+
+        def subscribe(self, *args, **kwargs):
+            ...
+
+    nspace["get_ipython"] = IPDummy(nspace)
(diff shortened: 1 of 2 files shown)


dmgav:
I think we should copy those changes to the latest version of QueueServer which I will also post as a draft.

Another question: should we use RE and db instantiated in nslsii? We can probably discard db, but what about RE? Is there any reason to keep it?

bluesky_queueserver/manager/profile_ops.py
```diff
@@ -49,10 +49,33 @@ def load_profile_collection(path):
...
+        def __init__(self, user_ns):
+            self.user_ns = user_ns
+
+        def __call__(self):
+            return self
+
+        def register_magics(self, *args, **kwargs):
+            ...
+
+        def magics(
```
mrakitin, line 64: This seems to be an incomplete method.

dmgav: As I understand it's not a working code, but work in progress. There are some fundamental issues that needs to be resolved before it works.

dmgav:
One of the problems that I see is that we can not modify `nspace` from within a module executed by `runpy` by passing reference to `nspace` to this module as an entry to `nspace` dictionary.

bluesky_queueserver/manager/profile_ops.py
tacaswell: :+1:

tacaswell:
I think we either need to do one of:

 a) monkey patch `IPython` with the above dummy
 b) monkey patch nslsii to make `configure_base` a mostly empty function (suspect we should just keep the imports)
 c) change nslsii upstream so that if `get_ipython() is None` then do not do the things that need it
 d) make `configure_base` take `get_ipython` as an (optional) input

dmgav:
The feature is fully implemented in https://github.com/bluesky/bluesky-queueserver/pull/57

PR #58, opened 2020-10-02
Support for local imports in loaded beamline profiles

dmgav:
Added support for relative imports in loaded beamline profiles: path to original location of beamline profile collection is added to `sys.path` before startup files are patched and loaded from a temporary directory. Unit tests are added.


tacaswell:
:+1: my only quibble is that I would call these "local imports" not relative as relative makes me think of `from . import foo` which only works in modules vs this which is relying on the profile directory implicitly being in the path when IPython loads them.

tacaswell (approved):
@dmgav can push the green button if happy with the two commits I added.

dmgav:
@tacaswell Thanks.

mrakitin (commented):
Read it after it was merged. Looks good.

PR #59, opened 2020-10-05
FIX: add a lock around zmq access

tacaswell:
Serialize access to zmq socket from web requests


cryos (approved):
This fixes the failures I was seeing in my local development environment.

mrakitin (approved):
This worked for me very well!

dmgav:
The problem is that I rewrote this code substantially. My fix is exactly the same, but there are more changes, e.g. the socket is automatically reopened in case of failure.

tacaswell:
Closed because @dmgav has fixed this else where.

PR #60, opened 2020-10-05
Add uvicorn logging config file

mrakitin:
This customized logging config is useful to add timestamps to the logs produced by `uvicorn`.

To be used as:
```bash
$ uvicorn bluesky_queueserver.server.server:app --host localhost --port 8080 --log-config uvicorn-logging.yaml
```

The PR https://github.com/encode/uvicorn/pull/799 adds support for the `.yml` files.


mrakitin:
Not sure if we want to ship it with source distributions or wheels, but we may need to update/extend the docs and tests.
@dmgav, @tacaswell, @cryos, thoughts?

dmgav:
Where does .yaml file have to reside? We could probably enclose the file (make a directory with sample config files) and write the instructions on how to manually install it if needed. Why would we need to test built-in `uvicorn` logging?

mrakitin:
We can have a separate `config` directory with such files. It may also be useful to ship it via a pip-installable package so that the server can be started with the desired log formatting.

Testing with `uvicorn` logging configured via a file can be useful to catch the potential errors if such a log file is not compatible with newer versions of `uvicorn` (currently it's extracted from the default logging configuration of `uvicorn`, but can potentially change). Just to be clear - we don't want to test the `uvicorn` itself, but it's config stored locally in our repo against the latest version of the `uvicorn`.

mrakitin:
Just discovered that [`fastapi[all]` requirements](https://github.com/tiangolo/fastapi/blob/e77ea635777d2494690ba3eb62bd005b9edeefde/pyproject.toml#L94) conflict with the new version of [`uvicorn` v0.12.0](https://github.com/encode/uvicorn/releases/tag/0.12.0), which provides the new feature for supporting external `.yaml` log config files. We'll have to wait until `fastapi` is compatible with it. I'll keep this PR as a draft until then.

mrakitin:
Closing as it's not actual for now. Can be revisited/reopened in the future if needed.

PR #64, opened 2020-10-08
Validation of experimental plans.

dmgav:
Implemented simple validation of experimental plans based on descriptions of plan parameters automatically extracted from beamline profile collections with the help of ``inspect`` module. The validation procedure catches the cases of incorrect number of args or incorrect kwargs. More data on the plan arguments including the lists of allowed Ophyd devices for each plan argument is needed to provide accurate validation.


dmgav:
This is a small PR. Most changes are due to the new autogenerated `.yaml` file.

mrakitin (approved):
Looks good to me, I don't see any issue with that, thanks!

dmgav:
I think we should just merge it. It is very likely that the function will have to be rewritten in the next step once it becomes clear what the next step is.

PR #65, opened 2020-10-09
CI: minor tweaks to .travis.yml file

mrakitin:
This is mostly a "cosmetic" PR, but I don't think the following line ever worked as the variable wasn't expanded:
```bash
   if [ BUILD_DOCS == 'true' ]; then
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,11 @@ jobs:
 
 before_install:
   - |
-    # set -e
+    set -e
     if [ "${FLAKE_8}" == "true" ]; then
         pip install flake8
         flake8 .
-        let res=$?
+        res=$?
         echo "The project code was verified with 'flake8'. Exit code ($res)."
         exit $res
     fi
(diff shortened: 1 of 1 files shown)


mrakitin:
I think I also fixed an issue with the opposite exit code introduced by `let`:
```bash
$ let "res=0"
$ echo $?
1
```

After fixing that, `sed -e` works as expected.

I also separated the fast jobs (style checks and docs building) from the heavy jobs with `pytest`. This way we'll know faster if something is wrong with the style or docs.

dmgav:
My only concern is if Travis exits with the right code if `flake8` fails. I remember I had to experiment with the sequence of commands until I got it working. Did you check if failing `flak8` or doc building actually fail the build?

tacaswell:
We might also consider switching to GH action for flake8.  We switched on mpl and now get nice in-line message when things are wrong: https://github.com/matplotlib/matplotlib/blob/master/.github/workflows/reviewdog.yml

mrakitin:
> We might also consider switching to GH action for flake8. We switched on mpl and now get nice in-line message when things are wrong: https://github.com/matplotlib/matplotlib/blob/master/.github/workflows/reviewdog.yml

Good point! I'll try to finish this PR with the current approach as it touches some other functionality. Then we can try the shiny thing.

dmgav:
Bump.

mrakitin:
@dmgav, I think it's ready to go. Feel free to apply the squash-merge for the PR to eliminate not-so-useful commits.

PR #66, opened 2020-10-14
First implementation of user groups

dmgav:
- User groups are defined in a YAML file in a profile collections directory. User group permissions can be also loaded from a different location: path can be specified in the command-line arguments of the RE Manager. File name can also be set if the default file name is not satisfactory. YAML file is manually editable. The loaded dictionary is checked using `jsonschema` to detect possible errors.

- A subsets of allowed plans and devices is obtained from the existing sets by applying rules for 'allowed' plans and devices and 'forbidden' plans and devices. Only plans and devices that satisfy at least one of the rules for 'allowed' plans and do not satisfy any of the 'forbidden' plans and devices are selected. Regular expressions are used to formulate the rules: there is no need to explicitly list the devices and plans.

- Previously defined 'allowed' plans and devices are renamed into 'existing' plans and devices. The YAML file was also renamed.

- Separate dictionaries of allowed plans and devices are created for each defined user group.

- Validation of added plans is performed using the dictionary of allowed plans for the user who is submitting the plan. User name and group are sent as additional (now required) parameters of RE Manager API method `queue_plan_add`. (Since the server is currently not supporting logins, the hard coded user name and group are sent to RE Manager. User name/group should not be sent as part of REST API call. CLI tool is also providing fixed user name and group).


bluesky_queueserver/manager/profile_ops.py
```diff
@@ -556,17 +559,189 @@ def load_list_of_plans_and_devices(path_to_file=None):
...
+    items_selected = {}
+    for item in item_dict:
+        select_item = False
+        if allow_patterns:
+            if allow_patterns[0] is None:
+                select_item = True
+            else:
+                for pattern in allow_patterns:
+                    if re.search(pattern, item):
+                        select_item = True
```
mrakitin, line 679: Should we break the for-loop here in case it's found?

bluesky_queueserver/manager/profile_ops.py
```diff
@@ -556,17 +559,189 @@ def load_list_of_plans_and_devices(path_to_file=None):
...
+                        select_item = False
+        if select_item:
+            items_selected[item] = copy.deepcopy(item_dict[item])
+
+    return items_selected
+
+
+def load_allowed_plans_and_devices(path_existing_plans_and_devices=None, path_user_group_permissions=None):
+    """
+    Generate dictionaries of allowed plans and devices for each user group. If there is no user
```
mrakitin, line 693:
```suggestion
    Generate dictionaries of allowed plans and devices for each user group. If there are no user
```

bluesky_queueserver/manager/start_manager.py
```diff
@@ -192,13 +192,22 @@ def start_manager():
...
-        "'allowed_plans_and_devices.yaml' is used.",
+        "'existing_plans_and_devices.yaml' is used.",
+    )
+    parser.add_argument(
+        "--user_group_permissions",
+        dest="user_group_permissions_path",
+        type=str,
+        help="Path to file that contains lists of plans and devices available to users. "
+        "The path may be a relative path to the profile collection directory. "
+        "If the path is directory, then the default file name "
```
mrakitin, line 209:
```suggestion
        "If the path is a directory, then the default file name "
```

bluesky_queueserver/server/server.py
```diff
@@ -9,6 +9,11 @@
 logger = logging.getLogger(__name__)
 
 
+# Login and authentication are not implemented, but some API methods require
+#   login data. So for now we set up fixed user name and group
+_login_data = {"user": "John Doe", "user_group": "admin"}
```
mrakitin, line 14: Can these parameters be a part of the kwargs for the `__init__` method below? Otherwise, it's hard to pass an arbitraty name if needed.

mrakitin (commented):
Looks good to me. I have a couple of suggestions.

bluesky_queueserver/server/server.py
dmgav: The name and user group should be determined at the time of user login. Since login is not implemented in the demo, we will use fixed user and group names.

bluesky_queueserver/manager/profile_ops.py
dmgav: Yes, thanks. Done.

PR #67, opened 2020-10-15
Separate tests for RE Manager ZMQ API (only a few tests for now).

dmgav:
Set up framework for direct testing of RE Manager ZMQ API. To this point API were tested indirectly by either calling `qserver` CLI or using REST API. From this point it is likely that CLI commands and REST APIs will start diverging from ZMQ API used for internal communication between the server and RE manager, so it would be beneficial to separate the tests as well. 

Only a few tests are added so far. There are also some minor changes to the code (mostly error messages).

### DIFF
--- a/bluesky_queueserver/manager/tests/_common.py
+++ b/bluesky_queueserver/manager/tests/_common.py
@@ -5,10 +5,14 @@ import pytest
 import subprocess
 import asyncio
 import time as ttime
+import zmq
 
 from bluesky_queueserver.manager.profile_ops import get_default_profile_collection_dir
-from bluesky_queueserver.manager.qserver_cli import CliClient
 from bluesky_queueserver.manager.plan_queue_ops import PlanQueueOperations
+
+import logging
+
+logger = logging.Logger(__name__)
 
 
 def copy_default_profile_collection(tmp_path, *, copy_yaml=True):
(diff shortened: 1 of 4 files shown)


bluesky_queueserver/manager/tests/_common.py
```diff
@@ -83,28 +87,133 @@ def patch_first_startup_file_undo(pc_path):
         os.rename(fln_tmp, fln)
 
 
+class TinyZmqClient:
+    """
+    This is a simplified version of ZMQ Client, which is very similar to the client
+    used in ``qserver_cli``. The server does not perform any verification of input
+    and output data and intended for testing 0MQ API of RE Manager.
+    """
```
mrakitin, line 95: Would it make sense to centralize all 0MQ clients in one place? It feels like a duplicate functionality reimplemented in a few places (http server, qserver_cli, etc.)

mrakitin (commented):
Looks good once the tests are passing. I have thought about the long-term maintenance of 0MQ clients.

bluesky_queueserver/manager/tests/_common.py
dmgav: It's a good question. I think that with time functionality of the calls made via server will change and the tests will change. I don't think that duplicate testing is a problem right now. Also the functioning of the CLI and REST API has to be checked with different combinations of parameters as well. I think that CLI interface has to be changed also to be more user friendly, so the tests will change.

mrakitin: OK, let's keep it in mind in the future, and keep it as is for now.

mrakitin (approved):

PR #72, opened 2020-10-18
More accurate validation of submitted plans.

dmgav:
Contributions:

- Procedure for validating plans submitted to QueueServer. Validation is performed before plans are sent to RE Worker process or even before the worker process is started. The procedure performs validation of plans based on stored plan annotations which are prepared in advance, so it does not require access to profile collection.

- Comprehensive testing of the validation procedure.

The plan validation procedure was developed based on ideas discussed with @tacaswell .


mrakitin (commented):
It looks like it overlapped with #68. Let's merge that PR and then rebase on `main` for easier reading of this PR.

PR #75, opened 2020-10-21
ZMQ client API

dmgav:
Contributions: 

- Moved all ZMQ client code to `comms.py`. Class `ZMQCommSendAsync` is used by the web server and function `zmq_single_request` (useful for episodic requests to RE Manager) is used by the CLI tool and unit tests. The API may also be used by third party software if needed.

- Fixed an issue in `server.py` (potentially two event loops).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ before_script:
             echo "Not attempting to run doctr due to the docs failed to build"
         else
             pip install doctr
-            doctr deploy --built-docs docs/build/html .
+            doctr deploy --built-docs docs/build/html . --branch-whitelist main
             res=$?
             echo "Docs deployment with 'doctr' finished with the exit code ($res)."
         fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,9 @@ jobs:
     # As of 2020-10-08, Python 3.9 distribution is not available in TravisCI's repos yet.
     # - python: 3.9
 
+  allow_failures:
+    - name: "Build/deploy docs"
+
 before_install:
   - |
     if [ "${FLAKE_8}" == "true" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,12 +65,14 @@ before_script:
         echo "Docs build finished with the exit code ($res)."
 
         if [ $res -gt 0 ]; then
-            echo "Not attempting to run doctr due to the docs failed to build"
+            echo "Not attempting to run doctr."
         else
-            pip install doctr
-            doctr deploy --built-docs docs/build/html . --branch-whitelist main
-            res=$?
-            echo "Docs deployment with 'doctr' finished with the exit code ($res)."
+            if [ "${TRAVIS_REPO_SLUG}" == "bluesky/bluesky-queueserver" ]; then
+                pip install doctr
+                doctr deploy --built-docs docs/build/html . --branch-whitelist main
+                res=$?
+                echo "Docs deployment with 'doctr' finished with the exit code ($res)."
+            fi
         fi
         exit $res
     fi
--- a/bluesky_queueserver/manager/comms.py
+++ b/bluesky_queueserver/manager/comms.py
@@ -426,10 +426,9 @@ class ZMQCommSendAsync:
     ----------
     loop: asyncio loop
         Current event loop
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -65,7 +68,7 @@ before_script:
             echo "Not attempting to run doctr due to the docs failed to build"
         else
             pip install doctr
-            doctr deploy --built-docs docs/build/html .
+            doctr deploy --built-docs docs/build/html . --branch-whitelist main
```
mrakitin, line 71: After looking at the corresponding code of [`doctr`](https://github.com/drdoctr/doctr/blob/1e5be24700ca5b07fa1603ccc4b89977dbd1cd78/doctr/__main__.py#L306), this looks like the right solution 👍

bluesky_queueserver/manager/comms.py
```diff
@@ -3,6 +3,8 @@
 import json
 import asyncio
 import uuid
+import zmq
+import zmq.asyncio
```
mrakitin, line 7: This additional import should help with tests on Mac. Some relevant info: https://github.com/zeromq/pyzmq/issues/1314#issuecomment-512753984.

bluesky_queueserver/manager/comms.py
```diff
@@ -412,3 +414,211 @@ def _pipe_receive(self):
...
+        returns error message or raises ``CommTimeoutError`` exception depending on
+        the setting of ``raise_timeout_exceptions`` property.
+
+        Parameters
+        ----------
+        method: str
+            Name of the method to be invoked on the server. The method must be supported
+            by the server.
+        params: dict or None
+            Dictionary of parameters passed to the method. If ``None`` then empty dictionar
```
mrakitin, line 541:
```suggestion
            Dictionary of parameters passed to the method. If ``None``, then an empty dictionary
```

bluesky_queueserver/server/server.py
```diff
@@ -1,112 +1,36 @@
...
+async def startup_event():
+    global zmq_to_manager
+    # ZMQCommSendAsync should be created from the event loop of FastAPI server.
+    zmq_to_manager = ZMQCommSendAsync()
+
+
+@app.on_event("shutdown")
+def shutdown_event():
+    global zmq_to_manager
+    zmq_to_manager.close()
```
mrakitin, line 33: Interesting solutions.

mrakitin (commented):
Looks good, I like the single place for the ZMQ comm. I have a very minor suggestion.

.travis.yml
```diff
@@ -62,12 +65,14 @@ before_script:
         echo "Docs build finished with the exit code ($res)."
 
         if [ $res -gt 0 ]; then
-            echo "Not attempting to run doctr due to the docs failed to build"
+            echo "Not attempting to run doctr."
         else
             pip install doctr
```
mrakitin, line 70: This install step can go inside the condition too.

dmgav: Done

mrakitin (approved):

dmgav:
Merging.

bluesky_queueserver/manager/comms.py
```diff
@@ -412,3 +414,211 @@ def _pipe_receive(self):
...
+        msg = msg_received
+        msg_err = ""
+    except Exception as ex:
+        msg = None
+        msg_err = str(ex)
+
+    if msg_err:
+        logger.warning("Communication with RE Manager failed: %s", str(msg_err))
+
+    return msg, msg_err
```
danielballan, line 624: Why not raise this? This pattern is common in JS but strikes me as unusual in Python.

dmgav: There is no particular reason except that this function is a product of evolution and it was moved here almost without change. I think I will fix it when I make another pass over this code (this function is widely used in tests, so there is some editing involved). Also this function is odd, because it does a lot of unnecessary stuff just to send a single zmq request, so it should be used in production only in cases where requests are infrequent or single (as in the CLI tool). We may need to think about another API for use in event-driven applications (such as Qt applications) that would provide modes including blocking wait, polling and callbacks. I think that if we cover both synchronous or thread-based and asynchronous cases, there is better chance the API will be used.

PR #76, opened 2020-10-25
Extension of API for operations with the items of the plan queue. 

dmgav:
Contributions (extension of QServer API):

- RE Manager API `queue_plan_add`, `queue_plan_get` and `queue_plan_remove` now support addressing by position (item number, "front", "back") or by plan UID. `queue_plan_add` can add plan before or after the plan with specified UID.

- New API `queue_plan_move`, which moves the plan from its current position in the queue to a new position in the queue. Source and destination may be specified as a numeric position, "front", "back" or plan UID (UID of a plan to be moved and UID of a new 'neighbour' of the plan, the plan can be inserted before or after the 'neighbour'). The addressing can be mixed, e.g. a plan with a given UID can be moved to the back or the front of the queue.

- Support for the modified API in RE Manager, HTTP server and CLI tool.

- Full test coverage of the API including edge cases.


tacaswell:
My knee-jerk reaction is that the manager should only handle the absolute case of "put in front/behind of this uid" and push the position -> uid conversion off to the clients.  On one hand, any sort of persistent client (eg the web / gui frontend) should know enough to do this look up (and there is something to be said for doing abmigious thing that is at least consistent with what the user sees when they make the decision), but on the other hand I see how from the CLI you don't have that state to work from.

dmgav:
@tacaswell The idea was to have API that cover all possible use cases. First of all, when plans are addressed by position, the acceptable position values `front` and `back`, which always have absolute meaning. Also if elements are addressed using negative indices (counted from the back of the queue), then the addressing is reliable if the elements are pulled from the front of the queue. Addressing using positive indices is also reliable if the queue is not running. If multiple users are going to drag plans back and forth in the queue at the same time, then it is going to be a mess with any addressing scheme.

dmgav:
> (and there is something to be said for doing abmigious thing that is at least consistent with what the user sees when they make the decision),

Could you clarify what you mean by ambiguous thing?

tacaswell:
I mean if the user sees `[plan_1, plan_2, plan_3]` and they say "add 'plan_4' as the 2nd plan, then on the client side we can convert that to a uuid based request (to either be "before plan_2" or "after plan_1", does not matter which so long as we are consistent about it) on the client side.   My initial thought was that this would make things more predictable, but now that I type this out I am no longer sure that is true (due to the ambiguity of the choice between which side we pin to).

[sorry, typed this out yesterday and failed to hit "comment" :( )

cryos:
I agree with @tacaswell on predictability, adding a plan to the end of the queue is entirely clear. The client should be careful to add before/after a uid when position is specified based on what the client sees. It will need to handle error cases if for example you add a plan after `uid blah` and then that plan is removed from the queue as the queue is running and it processed them quicker than the request was sent. The client presents a view of the queue that may be out of date, and it should make calls that attempt to do as the user requested as much as possible and then error gracefully if that is not possible.

tacaswell:
There is also the idea of having an incrementing (or random) "version number" on the queue and reject any possibly ambiguous instruction if the queue has changed sense the information the client last got.  This maybe be getting too complicated for now though.

cryos:
That sounds too complicated, and if we could use UIDs and handle error conditions that may well be enough.

dmgav:
@cryos @tacaswell All errors are handled. All API will return `success=False` in case the operation can not be completed. Also all operations are 'atomic' - queue can not changed while the operation is in progress. So the request will be either rejected or executed correctly. Again, if multiple users are going to move elements back and forth in the queue, it will look like a mess with any addressing.

cryos:
@dmgav They might be handled on the server, I was talking about the client handling the error and telling the user what happened. This relies on the server correctly stating this operation cannot happen because UID blah was not in the queue, but then the client needs to offer the user options or at least inform them their plan was not added as UID blah was not in the queue. There is a collaboration between client and server as their respective views may be slightly out of sync. I doubt this will happen often, if simply adding to the end for example, which should be the default, we will be fine. But when moving plans up/down the stack, there may be times when the relative UID disappears between the client making the call and the server processing the call.

dmgav:
There is an error message returned as well. But if `success==False`, then you can consider that the plan was not added.

tacaswell:
I am :+1: on merging this so we can keep moving.  I think this is fundamentally correct as-is and we should iterate on it after it is merged.

tacaswell:
unless @gwbischof and @cryos have concerns that this is not _enough_ of an API to write the web client.

gwbischof (approved):
The API changes look like they will be able to support the features that we have in mind for the web client.

cryos:
Agreed, I was just clarifying my comments about errors in the client. I felt that there was enough in the server, but talking through the logic I think should like in the client. I think this looks good, and should be merged.

PR #78, opened 2020-10-29
Improvements in the code for loading beamline profile collections

dmgav:
Improved patching of startup files:

- Support the case when `get_ipython()` function is not imported (insert the patch to the beginning of the startup file).

- Support of multiple imports of `get_ipython()` (not used anywhere, but may be useful to support anyway).

- Stashing exceptions raised during loading of startup files (`runpy`) and reraising them in the main applications.

- Decorator `set_user_ns` that allows to access currently used global user namespace from any function in the profile collection modules when the profiles loaded and executed (both with/without IPython).  

The changes may be useful in adaptation of the existing beamline profile collection for working with QueueServer.


bluesky_queueserver/manager/annotation_decorator.py
```diff
@@ -412,9 +412,17 @@ def parameter_annotation_decorator(annotation):
...
 
     def function_wrap(func):
-        @functools.wraps(func)
-        def wrapper(*args, **kwargs):
-            return func(*args, **kwargs)
+        if inspect.isgeneratorfunction(func):
+
+            @functools.wraps(func)
+            def wrapper(*args, **kwargs):
+                yield from func(*args, **kwargs)
```
tacaswell, line 419:
```suggestion
                return (yield from func(*args, **kwargs))
```

bluesky_queueserver/manager/profile_ops.py
```diff
@@ -33,10 +36,12 @@ def get_default_profile_collection_dir():
 
 _patch1 = """
 
+import sys
 import logging
+from bluesky_queueserver.manager.profile_tools import global_user_namespace
 logger_patch = logging.Logger(__name__)
 
-__local_namespace = locals()
+global_user_namespace.set_user_namespace(user_ns=locals(), use_ipython=False)
```
tacaswell, line 44: do we want to start with `locals()` here or `{}`?

dmgav: I am not sure what do you mean by `{}`. The reasoning for this is that each time when we start `runpy` the references to the namespace is changing, so we need to set  the globally accessible reference to the new value (this is the reason the reference should not be stored by the objects in the plan). We do the patching only for startup files, so all the functions from external modules that are called from start up files can use the current global reference. Initially we used only `__local_namespace__` from the patched startup files.

tacaswell: I mean do we want to seed this from an empty dictionary or the set of things that we have imported up until now.

dmgav: Ok, I understand the question now. We need `locals()` here.

dmgav: It is seeded from an empty dictionary before loading the first startup file.

PR #81, opened 2020-11-06
Automated conversion (reduction) of plan descriptions to satisfy requirements of web client.

dmgav:
Two contributions:

- Improved the procedure of selection of allowed plans and devices. The `root` user group is now selecting allowed devices and plans that could potentially be made available for users of other groups. The devices and plans that are not selected by `root` may not be accessed by users of any group. The main purpose is to filter out 'junk' (outdated, unused or non-functional plans and devices that may exist in profile collection) or plans that are not supposed to be manually executed or passed as parameters.

- Conversion of plan descriptions for use by web client. Conversion is performed at the HTTP server. Currently the conversion is performed each time the list of available plans is requested.
The simplified plan description contains the following fields:

```
        {
            "name": "<plan_name>",  # required
            "description": "<plan_description>",  # optional
            "parameters": {
                "name": "<parameter_name>",  # required
                "description": "<parameter_description>",  # optional
                "kind": <parameter_kind>,  # e.g. POSITIONAL_OR_KEYWORD, optional
                "type": <parameter_type>,  # "int", "float" or "str", optional
                "enum": ["..", "..", ".."],  # List of enum values (strings) if the parameter is enum, optional
                "is_list": True,  # True or False, optional
                "default": <default_value>,  # Default value (if available), optional
                "is_optional": True,  # True or False, indicates if the parameter is optional, optional
                "min": <min_parameter_value>,  # optional
                "max": <max_parameter_value>,  # optional
                "step": <step_value>,  # optional
            },
        }
``` 
Missing ``type`` means that type is unknown. If the parameter is enum, then ``type`` is
always ``str``. Missing ``is_list`` means that it is unknown if the parameter accepts a list.
The parameter is optional (``is_optional is True``) if the default value is specified.
Step value may be specified in the custom parameter annotation if the value is discrete
and changes from ``min`` to ``max`` value with fixed ``step``.

### DIFF
--- a/bluesky_queueserver/server/conversions.py
+++ b/bluesky_queueserver/server/conversions.py
@@ -12,7 +12,7 @@ def filter_plan_descriptions(plans_source):
     The limitation on processed types is based on potential capabilities of the web client.
     Complex type hints may also be processed correctly if the result could be expressed
     in terms of the supported types. The performance of the conversion function must
-    be checked with the plans using complex types before deployement.
+    be checked with the plans using complex types before deployment.
 
     The converted plan parameters represent the following dictionary:
 
--- a/bluesky_queueserver/server/conversions.py
+++ b/bluesky_queueserver/server/conversions.py
@@ -3,11 +3,44 @@ import re
 
 def filter_plan_descriptions(plans_source):
     """
-    Filter the descriptions of the allowed plan to make them easier to use by the web client.
-    There are some assumptions that are made when reducing complex types that are used
-    for parameter validation by the RE Manager to simple types. Those assumptions may not
-    always be valid, therefore operation of the function should always be validated when
-    working with the plans that require complex types.
+    Convert descriptions of the allowed plans to satisfy requirements of the web client.
+    The conversion accurately supports only simple data types: ``int``, ``float``, ``str``,
+    lists of ints, floats or strings (expressed as ``List[int]`` or ``typing.List[int]``),
+    enums and lists of enums. Enum can be specified in custom description of plan
+    parameter with enum values which are string or names of plans or devices.
+
+    The limitation on processed types is based on potential capabilities of the web client.
+    Complex type hints may also be processed correctly if the result could be expressed
+    in terms of the supported types. The performance of the conversion function must
+    be checked with the plans using complex types before deployement.
+
+    The converted plan parameters represent the following dictionary:
+
+    .. code-block:: python
+
+        {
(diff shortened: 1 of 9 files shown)


bluesky_queueserver/server/conversions.py
```diff
@@ -0,0 +1,144 @@
...
+    Convert descriptions of the allowed plans to satisfy requirements of the web client.
+    The conversion accurately supports only simple data types: ``int``, ``float``, ``str``,
+    lists of ints, floats or strings (expressed as ``List[int]`` or ``typing.List[int]``),
+    enums and lists of enums. Enum can be specified in custom description of plan
+    parameter with enum values which are string or names of plans or devices.
+
+    The limitation on processed types is based on potential capabilities of the web client.
+    Complex type hints may also be processed correctly if the result could be expressed
+    in terms of the supported types. The performance of the conversion function must
+    be checked with the plans using complex types before deployement.
```
mrakitin, line 15:
```suggestion
    be checked with the plans using complex types before deployment.
```

bluesky_queueserver/server/conversions.py
```diff
@@ -0,0 +1,144 @@
...
+    -------
+    dict
+        Dictionary of plans with reduced set of parameters.
+    """
+
+    plans_filtered = {}
+    for p_name, p_items in plans_source.items():
+
+        plan = dict()
+        # Plan name - mandatory (actually it is always equals 'pname')
```
mrakitin, line 60:
```suggestion
        # Plan name - mandatory (actually it always equals 'pname')
```

mrakitin (approved):
Looks good to me. Spotted a couple of places in docstrings, suggested corrections.

gwbischof:
> If the parameter is enum, then type is
always str.

Is the reverse true? If the type is string then it is an enum?

I think it would be useful to know if the type is enum or string, is there a way that I can tell the difference?

dmgav:
@gwbischof No type `str` does not automatically mean that it's `enum`, but `enum` is always `str`. You can look at it this way: `enum` is a list of values, each value always has type `str`. Since you pick one of the values and pass it as a parameter, the parameter is also has type `str`. So there is no contradiction here. I guess the right way to interpret it is to check `enum` is specified and display a combo box for selecting the value from the list if it is specified. Type `str` without `enum` means that you can attempt to send any string.

gwbischof (approved):

mrakitin (approved):

PR #83, opened 2020-11-09
CI: allow build to fail while deploying documentation in case of collision

dmgav:
Changes in this PR allow build to fail in case of collision, which can happen if several builds attempt to push changes to `gh_pages` branch simultaneously. Such event is extremely unlikely if the script `deploy_docs` is used as intended since it requires multiple PRs being merged simultaneously. The change includes replacement of `push --force-with-lease` by `push`. 

Additional changes to the script include multiple `echo` statements helpful for easier navigation in the script output.

### DIFF
--- a/.github/workflows/deploy_docs.sh
+++ b/.github/workflows/deploy_docs.sh
@@ -4,6 +4,8 @@
 #   e.g BUILD_DIR="docs/build/html"
 # GITHUB_TOKEN should be also set (e.g. GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }})
 
+echo "====================================================================="
+echo "SCRIPT: PUBLISHING DOCUMENTATION TO GITHUB PAGES"
 set -eu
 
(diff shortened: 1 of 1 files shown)


danielballan (approved):

.github/workflows/deploy_docs.sh
```diff
@@ -12,40 +14,58 @@ TARGET_BRANCH="gh-pages"
...
 git diff-index --quiet HEAD || git commit -am "Deployed Docs"
 echo "Changes were committed."
 
-git remote set-url "$REMOTE_NAME" "$REPO_URI" # includes access token
-git push --force-with-lease "$REMOTE_NAME" "$TARGET_BRANCH"
+echo "====================================================================="
+echo "Pushing changes to the repository ..."
+git remote set-url "$REMOTE_NAME" "$REPO_URI"  # includes access token
+# Allow 'push' to fail in case of collision with another build
+git push "$REMOTE_NAME" "$TARGET_BRANCH"
```
tacaswell, line 70: Don't we need some level of forcing here because on L63 we are amending the current commit (which means we always only have 1 commit on the gh-pages branch).

PR #89, opened 2020-11-20
FIX: reject create/close environment requests if already underway (2nd attempt)

developingAlex:
Make use of the RE Manager MState enum to indicate when the
manager is already underway in the creation or closing of a
RE Worker environment.

Previously the manager would get into a locked up state if
a request to create an environment was received while it was
already mid-way through the act of creating an environment.

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -149,7 +149,7 @@ class RunEngineManager(Process):
         elif self._manager_state is MState.CREATING_ENVIRONMENT:
             accepted, msg = False, "Manager is already in the process of creating the RE Worker environment."
         elif self._manager_state is not MState.IDLE:
-            accepted, msg = False, f"Manager state was {self._manager_state.value}"
+            accepted, msg = False, f"Manager state is not idle. Current state: {self._manager_state.value}"
         else:
             accepted, msg = True, ""
             self._manager_state = MState.CREATING_ENVIRONMENT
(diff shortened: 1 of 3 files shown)


dmgav:
@developingAlex Thank you for noticing the problem. We did not have time to go through minor details yet. I pushed two additional commits that includes additional checks inside the 'task' functions (to make sure that the manager is still in correct state, because in certain circumstances the state can change) and several unit tests.

I think the PR can be merged. @mrakitin

dmgav (approved):

bluesky_queueserver/manager/manager.py
```diff
@@ -145,21 +145,29 @@ async def _start_re_worker(self):
...
     async def _start_re_worker_task(self):
         """
         Creates worker process.
         """
+        # Repeat the checks, since it is important for the manager to be in the correct state.
         if self._environment_exists:
-            return False, "Rejected: environment already exists"
+            return False, "Rejected: RE Worker environment already exists"
+
+        if self._manager_state is not MState.IDLE and self._manager_state is not MState.CREATING_ENVIRONMENT:
```
mrakitin, line 167:
```suggestion
        if self._manager_state not in [MState.IDLE, MState.CREATING_ENVIRONMENT]:
```

bluesky_queueserver/manager/manager.py
```diff
@@ -205,11 +220,13 @@ async def _stop_re_worker_task(self):
...
         """
+        # Repeat the checks, since it is important for the manager to be in the correct state.
         if not self._environment_exists:
-            return False, "Environment does not exists"
+            self._manager_state = MState.IDLE
+            return False, "Environment does not exist"
 
-        if self._manager_state != MState.IDLE:
-            return False, "A plan is currently running"
+        if self._manager_state is not MState.IDLE and self._manager_state is not MState.CLOSING_ENVIRONMENT:
```
mrakitin, line 228:
```suggestion
        if self._manager_state not in [MState.IDLE, MState.CLOSING_ENVIRONMENT]:
```

mrakitin (approved):
Thanks, @developingAlex! The addition looks great to me as it allows better control of the state of the manager. I suggested a couple of small enhancements (optional).

developingAlex:
Hey @mrakitin and @dmgav , I only just saw your comments now, thanks for accepting and I like your additional commits very much, thanks!

PR #90, opened 2020-11-22
DOC: correct typo in flake8 command

developingAlex:
Hey please consider this somewhat of a tentative PR, I'm not sure if the documentation as is would have worked for some setups? But I found when I tried to run it that the slash was needed as I guess it was meant to be a path, so I assumed this was a typo?

**Also I gained a bit of experience in trying to run the unit tests** on my local machine that I felt wasn't really documented in the current README or CONTRIBUTING docs. **I'm happy to add those steps to the documentation** provided I didn't miss it already being documented or mentioned somewhere else already?
It wouldn't be much, just the steps to execute and the prerequisites (**like having a running redis server when running the tests, and http server**)

Regarding my experience having to manually start the redis service and http server, were the tests supposed to do this automatically for me? I wasn't sure but they didn't seem to.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -75,7 +75,7 @@ Ready to contribute? Here's how to set up `bluesky-queueserver` for local develo
 
 5. When you're done making changes, check that your changes pass flake8 and the tests, including testing other Python versions with tox::
 
-    $ flake8 bluesky_queueserver/tests
+    $ flake8
     $ python setup.py test
     $ tox
 
@@ -101,4 +101,3 @@ Before you submit a pull request, check that it meets these guidelines:
 3. The pull request should work for Python 2.7, 3.3, 3.4, 3.5 and for PyPy. Check
    https://travis-ci.org/tacaswell/bluesky-queueserver/pull_requests
    and make sure that the tests pass for all supported Python versions.
-
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -75,7 +75,7 @@ Ready to contribute? Here's how to set up `bluesky-queueserver` for local develo
 
 5. When you're done making changes, check that your changes pass flake8 and the tests, including testing other Python versions with tox::
 
-    $ flake8 bluesky_queueserver tests
+    $ flake8 bluesky_queueserver/tests
     $ python setup.py test
     $ tox
 


dmgav:
@developingAlex Thanks for pointing out one more issue. It looks like the file `CONTRIBUTING.rst` is autogenerated based on https://github.com/NSLS-II/scientific-python-cookiecutter/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/CONTRIBUTING.rst and not yet revised for this project. For this project we are using `black` for formatting and `flake8` to check style. `flake8` checks more than just formatting, so it is not redundant. So the correct sequence would be to run
```
$ black .
$ flake8
```
in the root directory of the project. The options for `black` and `flake8` are set in configuration files `.flake8` and `pyproject.toml`, so when you call them from the root directory, they are called with the same options as in CI. We are also testing the server only for Python 3.7 and 3.8. Also, as you mentioned, Redis and http server must be started before running tests.

In terms of documentation, we are currently maintaining only `README.txt`. The rest of the documentation (`docs` folder) was written before the work on the server was started, so it reflects only assumption about how the server should work, not how it was actually implemented. We are getting closer to starting documenting things properly.

CONTRIBUTING.rst
```diff
@@ -75,7 +75,7 @@ Ready to contribute? Here's how to set up `bluesky-queueserver` for local develo
 
 5. When you're done making changes, check that your changes pass flake8 and the tests, including testing other Python versions with tox::
 
-    $ flake8 bluesky_queueserver tests
+    $ flake8 bluesky_queueserver/tests
```
mrakitin, line 78: This file came from the templates via https://github.com/NSLS-II/scientific-python-cookiecutter. I think, historically some projects had the `tests` directory on the same level as the library with modules (`bluesky_queueserver/` in this case). As the directories with the tests are a part of the library now, this can be as simple as:
```suggestion
    $ flake8
```
There is a file `.flake8` in the root of the repo, which "controls" the flake8's behavior and operates the same way in [CI](https://github.com/bluesky/bluesky-queueserver/blob/e3a0fa0ab98ec6eb31bc7b1466b6b07bfa63badf/.github/workflows/flake8.yml#L21).

mrakitin (commented):
I think the suggested fix should make it better. Once merged, we will reconsider if it's worth to keep the file in the repo as is, or have it as a redirection to another resource.

mrakitin (approved):

mrakitin:
Merging for now. We will revisit this file later.

PR #91, opened 2020-11-23
CI: reliability of unit tests

dmgav:
The PR contains a number of small tweaks that make execution of unit tests more reliable.

PR #92, opened 2020-11-23
ENH: load reference to 'db' into RE namespace

dmgav:
Fixes the issue discovered during testing of the QueueServer. In the default mode, QueueServer is removing references to Run Engine (`RE`) and Data Broker (`db`) from RE namespace and generates a new instance of Run Engine. If the name of configuration file is provided as the value of one of CLI parameters (e.g. `start-re-manager --databroker_config "bmm"`), then a new instance of `db` is created and Run Engine is subscribed to it. A simple modification to the code in this PR includes the reference to `db` into RE namespaces so that it becomes accessible from the executed plans.

### DIFF
--- a/bluesky_queueserver/manager/worker.py
+++ b/bluesky_queueserver/manager/worker.py
@@ -565,6 +565,8 @@ class RunEngineWorker(Process):
                             from databroker import Broker
 
                             self._db = Broker.named(config_name)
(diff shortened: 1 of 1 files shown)


mrakitin:
Would it make sense to rebase this PR on `main` as some similar updates were just merged via #91?

mrakitin (approved):

PR #93, opened 2020-11-24
ENH: implementation of an option to add 'queue_stop' instructions to the plan queue

dmgav:
The code that implements an option to add `queue_stop` instructions to the plan queue. Once the instruction is encountered during plan execution, the queue execution is stopped. Execution may be resumed by sending `queue_start` request to the manager. The instructions are treated similarly to plans: they can be added to the queue, removed from the queue or moved within the queue. Instructions are assigned `plan_uid` and can be accessed by position and UID (the same way as plans).

Currently an instruction can be added to the queue by sending

```
"instruction": {"action": "queue_stop"}
```
instead of `"plan": {...}` as an argument of `queue_plan_add`. The option of adding instructions is not yet implemented in CLI tool.

The option of adding instructions to the queue makes current naming (e.g. `queue_plan_add`, `queue_plan_remove`, `plan_uid`) misleading. It may be a good idea to rename the API that are used to request operations on the queue and data fields reporting the state of the queue to use the word `item` instead of `plan` (e.g. `queue_item_add`, `queue_item_remove`, `item_uid`).

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -434,39 +434,61 @@ class RunEngineManager(Process):
             logger.info(err_msg)
 
         else:
-            # Reset RE environment (worker)
-            success, err_msg = await self._worker_command_reset_worker()
-            if not success:
-                self._manager_state = MState.IDLE
-                err_msg = f"Failed to reset RE Worker: {err_msg}"
-                logger.error(err_msg)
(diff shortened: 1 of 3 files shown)


mrakitin (approved):
It's a good addition, I see no issues with it.

dmgav:
Thanks @mrakitin. I'll start working on renaming things.

PR #94, opened 2020-11-26
Renaming of API to refer plan queue items as 'items' instead  of 'plans'.

dmgav:
The changes in this PR include renaming of a number of API (both 0MQ and REST) to reflect the expansion of functionality of the plan queue. Since the plan queue may now include instructions (currently only instructions that stops execution of the queue is implemented), it is more consistent to name queue items as 'items', not as 'plans'. The PR includes only trivial changes: there are no changes to functionality of the Queue Server.

**NOTE**: In existing installation of QueueServer, history may contain 'old' items that have `plan_uid` instead of `item_uid`. Clear history by running `qserver -c history_clear` to clear the history to ensure consistency of data returned by `/history/get` API.

The following API were renamed:

-  `queue_plan_add` renamed to `queue_item_add`,

-  `queue_plan_remove` renamed to `queue_item_remove`,

-  `queue_plan_get` renamed to `queue_item_get`,

-  `queue_plan_move` renamed to `queue_item_move`,

- `/queue/plan/add' renamed to `/queue/item/add`,

- `/queue/plan/add' renamed to `/queue/item/add`,

- `/queue/plan/remove' renamed to `/queue/item/remove`,

- `/queue/plan/get' renamed to `/queue/item/get`,

- `/queue/plan/move' renamed to `/queue/item/move`.

Additionally `plan_uid` is renamed to `item_uid`. The UID is assigned to each queue item and does not have meaning outside the queue. It has no relationship with Run UID.

The following status fields (returned following `status` request) were renamed:

- `plans_in_queue` renamed to `items_in_queue`,

- `plans_in_history` renamed to `items_in_history`,

- `running_plan_uid` renamed to `running_item_uid`.

### DIFF
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -32,7 +32,7 @@ class PlanQueueOperations:
         qsize = await pq.get_queue_size()
 
         # Read the queue (as a list)
-        queue = await pq.get_plan_queue()
+        queue = await pq.get_queue()
 
         # Start the first plan (This doesn't actually execute the plan. It is just for bookkeeping.)
         plan = await pq.set_next_plan_as_running()
@@ -81,7 +81,7 @@ class PlanQueueOperations:
         """
         Delete all the invalid queue entries (there could be some entries from failed unit tests).
         """
-        pq = await self._get_plan_queue()
+        pq = await self._get_queue()
 
         def verify_plan(plan):
             # The criteria may be changed.
@@ -182,7 +182,7 @@ class PlanQueueOperations:
         IndexError
             No plan is found.
         """
-        queue = await self._get_plan_queue()
+        queue = await self._get_queue()
         for n, plan in enumerate(queue):
             if plan["plan_uid"] == uid:
                 return n
@@ -238,7 +238,7 @@ class PlanQueueOperations:
         """
         Initialize ``self._uid_dict`` with UIDs extracted from the plans in the queue.
         """
-        pq = await self._get_plan_queue()
+        pq = await self._get_queue()
         self._uid_dict_clear()
         # Go over all plans in the queue
         for plan in pq:
@@ -327,26 +327,26 @@ class PlanQueueOperations:
         async with self._lock:
             return await self._get_queue_size()
 
-    async def _get_plan_queue(self):
+    async def _get_queue(self):
         """
-        See ``self.get_plan_queue()`` method.
+        See ``self.get_queue()`` method.
         """
         all_plans_json = await self._r_pool.lrange(self._name_plan_queue, 0, -1)
         return [json.loads(_) for _ in all_plans_json]
 
-    async def get_plan_queue(self):
+    async def get_queue(self):
         """
-        Get the list of all plans in the queue. The first element of the list is the first
-        plan in the queue.
+        Get the list of all items in the queue. The first element of the list is the first
+        item in the queue.
 
         Returns
         -------
         list(dict)
-            The list of plans in the queue. Each plan is represented as a dictionary.
+            The list of items in the queue. Each item is represented as a dictionary.
             Empty list is returned if the queue is empty.
         """
         async with self._lock:
-            return await self._get_plan_queue()
+            return await self._get_queue()
 
     async def _get_plan(self, *, pos=None, uid=None):
         """
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -439,9 +439,9 @@ class PlanQueueOperations:
                 f"The number of removed plans is {n_rem_plans}. One plans is expected to be removed."
             )
 
-    async def _pop_plan_from_queue(self, *, pos=None, uid=None):
+    async def _pop_item_from_queue(self, *, pos=None, uid=None):
         """
-        See ``self._pop_plan_from_queue()`` method
+        See ``self._pop_item_from_queue()`` method
         """
 
         if (pos is not None) and (uid is not None):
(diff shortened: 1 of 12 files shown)


tacaswell:
Attn @developingAlex .

gwbischof (approved):

gwbischof (commented):
What about /plans/allowed should this be renamed also?

dmgav:
@gwbischof `/plans/allowed` returns a list of allowed plans, so it should contain the word `plans`.

gwbischof:
I think instructions should be thought of as special plans, instead of a new type of object. Future instructions might have parameters and a plan form on the GUI. From the GUI point of view, instructions and plans behave the same way.  It would be convenient to get all the plans and instructions from /plans/allowed, and submit them to the queue with `add`.  It would also be nice if the schema of the `allowed_instructions` also matched the `allowed_plans` schema. For that reason I think instructions should be thought of as special plans.

dmgav:
It is a difficult question. I think that since instructions are not plans, we should keep clear distinction between them. One of the goals of renaming in this PR is to increase the distinction. Placing plans and instructions in the same queue is not supposed to indicate that plans and instructions are similar, since they are processed separately and have different purpose. Queue item can be considered as a container with assigned Item UID and some metadata needed for queue operations. Each item can hold either plan or instruction. 

In addition to plans and instructions, there is an idea to implement the ability to execute scripts in RE environment (scripts should be predefined in the profile collection). This would introduce the 3rd item type. There is no support for this idea at this point.

gwbischof:
the corresponding update for the webclient: https://github.com/bluesky/bluesky-webclient/pull/40/files#diff-0d855dce2c4cec96a9abddfa2b2a8419c62e95b831b57c1d183f16b03a90f842L52

dmgav:
@tacaswell Should we merge it?

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -678,79 +677,79 @@ async def _move_plan(self, *, pos=None, uid=None, pos_dest=None, before_uid=None
...
+    async def _add_to_history(self, item):
         """
-        Add the plan to history.
+        Add an item to history.
 
         Parameters
         ----------
-        plan: dict
+        item: dict
             Plan represented as a dictionary of parameters. No verifications are performed
```
mrakitin, line 753:
```suggestion
            Item (plan) represented as a dictionary of parameters. No verifications are performed
```

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -759,16 +758,16 @@ async def _add_plan_to_history(self, plan):
...
         """
-        See ``self.get_plan_history_size()`` method.
+        See ``self.get_history_size()`` method.
         """
         return await self._r_pool.llen(self._name_plan_history)
 
-    async def get_plan_history_size(self):
+    async def get_history_size(self):
         """
         Get the number of items in the plan history.
```
mrakitin, line 772:
```suggestion
        Get the number of items in the items history.
```

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -778,97 +777,97 @@ async def get_plan_history_size(self):
...
+        See ``self.clear_history()`` method.
         """
-        while await self._get_plan_history_size():
+        while await self._get_history_size():
             await self._r_pool.rpop(self._name_plan_history)
 
-    async def clear_plan_history(self):
+    async def clear_history(self):
         """
         Remove all entries from the plan queue. Does not touch the running plan.
```
mrakitin, line 812:
```suggestion
        Remove all entries from the queue. Does not touch the running item.
```

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -778,97 +777,97 @@ async def get_plan_history_size(self):
...
         The plan may be pushed back into the queue if it is stopped.
```
mrakitin, line 813:
```suggestion
        The item may be pushed back into the queue if it is stopped.
```

bluesky_queueserver/manager/plan_queue_ops.py
```diff
@@ -915,8 +914,8 @@ async def set_processed_plan_as_stopped(self, exit_status):
         Returns
         -------
         dict
-            The plan added to the history including ``exit_status``. If another no plan is currently
-            running, then ``{}`` is returned.
+            The plan added to the history including ``exit_status``. If another item (plan)
```
mrakitin, line 917: Should the `plan` be replaced with `item` here?

bluesky_queueserver/manager/tests/test_plan_queue_ops.py
```diff
@@ -569,191 +569,191 @@ async def testing():
...
-        assert await pq.get_plan_queue_size() == 2
+            await pq.pop_item_from_queue(uid=plan_to_remove["item_uid"])
+        assert await pq.get_queue_size() == 2
 
         # Attempt to remove the plan that is running. This should raise an exception.
-        await pq.set_next_plan_as_running()
-        assert await pq.get_plan_queue_size() == 1
+        await pq.set_next_item_as_running()
+        assert await pq.get_queue_size() == 1
         with pytest.raises(IndexError, match="Can not remove a plan which is currently running"):
```
mrakitin, line 643: Is the error message correct/expected?

mrakitin (commented):
Looks good to me. Spotted a couple of missed places in docstrings, and have a couple of questions.

bluesky_queueserver/manager/plan_queue_ops.py
dmgav: I think that we will still use 'plan queue' and 'plan history'. They are mostly be used for plans anyway, so the name makes sense. But they will contain 'items'.

bluesky_queueserver/manager/plan_queue_ops.py
dmgav: This will remain true only for plans unless we decide to support executing Python scripts, which will not happen.

bluesky_queueserver/manager/plan_queue_ops.py
dmgav: At this point the plan history contains only plans.

dmgav: I don't think it is critical, because currently only plans may have 'running' status. Execution of the one available instruction is instant. But let's generalize it.

mrakitin (approved):

dmgav:
@developingAlex We are merging this PR to move forward with the development and avoid significant rebase. If you have the code that is communicating with RE Manager, you will need to update API names. This should be a superficial change. We would like to be conservative about changing API later in the development, so it is preferable to implement all changes as early as possible before the Queue Server is deployed. Please let us know if you have any suggestions.

Issue #104, opened 2020-12-10
Inconsistencies in ZMQ API


dmgav:
Systematic review of the available ZMQ API showed some minor inconsistencies that should be fixed. The proposed changes are based on the following guidelines for consistency:

- Each request (with the exception of `status` request) is expected to return the parameters `success` and `msg`. `success` is a boolean (`True/False`) which indicates if the request was processed correctly (accepted) or failed (rejected). The parameter `msg` contains error message in case the request fails and the empty string otherwise.

- Plan queue items should be referred as `items`, not as `plans`.

**queue_get**
- Add return parameters `success` and `msg` (for consistency with the rest of API). 
- Rename `running_plan` to `running_item` (missed whan plan queue items were renamed from 'plans' to 'items').

**history_get**
- Add return parameters `success` and `msg` (for consistency with the rest of API). 

**history_clear**
- return parameter `msg` should be `""` if request is processed successfully.

**queue_item_get**, **queue_item_remove**, **queue_item_move**
- return parameter `plan` should be renamed to `item.

**queue_clear**
- return parameter `msg` should be `""` if request is processed successfully.

dmgav:
Addressed in https://github.com/bluesky/bluesky-queueserver/pull/105

PR #103, opened 2020-12-09
ENH: include UIDs of completed runs in plan execution results in plan history

dmgav:
This PR contains minor changes to RE Manager code that enable saving a list of Run UIDs in plan history. The example below demonstrates the contents of updated plan history after execution of two identical single-run plans. Each history item now contains a field named `result`, which is used to keep `exit_status` and `run_uids`. Note, that `run_uids` is a list that may contain multiple UIDs in case of multirun plans. Internally, the list of Run UID is Run Engine after completion of a plan.

**Minor API change:** `exit_status` was moved to `result` dictionary.

```
             [{'args': [['det1', 'det2']],
              'item_type': 'plan',
              'item_uid': '665f1af0-97e8-49c4-af10-a20f4b90ddff',
              'kwargs': {'delay': 1, 'num': 10},
              'name': 'count',
              'result': {'exit_status': 'completed',
                         'run_uids': ['8cea82eb-99f1-4fc7-b70d-f3c02b80ac92']},
              'user': 'qserver-cli',
              'user_group': 'admin'},
             {'args': [['det1', 'det2']],
              'item_type': 'plan',
              'item_uid': '8b9c4281-4737-4779-90c2-161a7bf74ed6',
              'kwargs': {'delay': 1, 'num': 10},
              'name': 'count',
              'result': {'exit_status': 'completed',
                         'run_uids': ['29ce9f55-e73e-4993-926e-d5546394394d']},
              'user': 'qserver-cli',
              'user_group': 'admin'}]
```

PR contains all commits from PR https://github.com/bluesky/bluesky-queueserver/pull/100

PR #105, opened 2020-12-10
Fix for minor inconsistencies in ZMQ API

dmgav:
The PR contains fixes for minor inconsistencies in ZMQ API. The PR implements changes suggested in Issue https://github.com/bluesky/bluesky-queueserver/issues/104. Changes are applied to the following API:

**queue_get**
- Add return parameters `success` and `msg` (for consistency with the rest of API). 
- Rename `running_plan` to `running_item` (missed whan plan queue items were renamed from 'plans' to 'items').

**history_get**
- Add return parameters `success` and `msg` (for consistency with the rest of API). 

**history_clear**
- return parameter `msg` should be `""` if request is processed successfully.

**queue_item_get**, **queue_item_remove**, **queue_item_move**
- return parameter `plan` should be renamed to `item.

**queue_clear**
- return parameter `msg` should be `""` if request is processed successfully.

The PR contains commit from PR https://github.com/bluesky/bluesky-queueserver/pull/100 and PR https://github.com/bluesky/bluesky-queueserver/pull/103

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -966,7 +966,7 @@ class RunEngineManager(Process):
         """
         logger.info("Clearing the queue")
         await self._plan_queue.clear_queue()
-        return {"success": True, "msg": "Plan queue is now empty."}
+        return {"success": True, "msg": ""}
 
     async def _history_get_handler(self, request):
         """
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -905,13 +905,13 @@ class RunEngineManager(Process):
             plan, msg = {}, ""
             pos = request.get("pos", None)
             uid = request.get("uid", None)
-            plan = await self._plan_queue.get_item(pos=pos, uid=uid)
+            item = await self._plan_queue.get_item(pos=pos, uid=uid)
             success = True
         except Exception as ex:
             success = False
             msg = f"Failed to get an item: {str(ex)}"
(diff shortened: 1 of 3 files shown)


bluesky_queueserver/manager/manager.py
```diff
@@ -944,29 +944,29 @@ async def _queue_item_move_handler(self, request):
...
+        return {"success": success, "msg": msg, "item": item, "qsize": qsize}
 
     async def _queue_clear_handler(self, request):
         """
         Remove all entries from the plan queue (does not affect currently executed run)
         """
         logger.info("Clearing the queue")
         await self._plan_queue.clear_queue()
-        return {"success": True, "msg": "Plan queue is now empty."}
+        return {"success": True, "msg": ""}
```
mrakitin, line 969: When is the `msg` supposed to be non-empty?

mrakitin (approved):
Looks good.

A single question: when is the `msg` supposed to be non-empty? In case of errors/exceptions?

bluesky_queueserver/manager/manager.py
dmgav: `msg` contains error message if request fails. For the requests that never fail `msg` is always empty string. The idea is that implementation of the API may change or API functionality may be expanded, so that there will be failing cases.

mrakitin: Over DM, @dmgav explained that if the `success=True`, then `msg` should always be empty.

Issue #114, opened 2020-12-18
Accept other ways of populating the worker's namespace


danielballan:
We currently use IPython profile startup scripts for most beamline configurations. IPython startup scripts are executed in a single namespace, sorted alphanumerically. To make that sorting explicit, we have adopted the _convention_ of prefixing them with a number, like `00-startup.py`.

This system is mostly a result of inertia, borne of copy/pasting the first working bluesky configuration (CSX) around the ring during NSLS-II's rapid scale-up. There are many downsides to our system. Two important ones are:

* Python scripts that begin with a digit are not importable as modules. This makes it difficult to build tests and documentation around them.
* Executing multiple scripts in one namespace may be convenient at small scales, but it confuses linters and modern editors (as they cannot see where variables are defined). This has led to serious bugs in practice. In one memorable incident at SRX, the critical `databroker.Broker` instance that was saving user data, `db`, was shadowed in a subsequent script and was therefore garbage collected and unsubscribed from the `RunEngine`.

The MVP of the worker was tuned to the reality that we use IPython profiles mostly named `profile_collection` and they tend to include scripts named like `[0-9][0-9]*.py`. (The worker skips any scripts that do not adhere to this convention even though IPython does not.)

https://github.com/bluesky/bluesky-queueserver/blob/994b858b0acadeb601aa94e5ec0c15fadc2ea93b/bluesky_queueserver/manager/profile_ops.py#L208-L210

This was a reasonable MVP. But I think we should pivot to work with other approaches and generally try to steer _away_ from IPython profiles in the future, given the downsides outlined above.

I suggest changing `start_re_worker` in the following ways:

- Provide a [mutually exclusive group of parameters](https://docs.python.org/3/library/argparse.html#mutual-exclusion) including:
    - a mode that accepts an importable module as in `-m my_pacakge.my_module`
    - a mode that accepts a script as in `--script path/to/script.py`
    - a mode that accepts an IPython profile as in `--profile PROFILE_NAME`
- The name `profile_collection` is just a convention (one with at least one exception within NSLS-II and possibly more elsewhere). I suggest changing `--profile-collection` to just `--profile PROFILE_NAME`, taking a profile name, and respecting [`IPYTHON_DIR`](https://ipython.readthedocs.io/en/stable/development/config.html) to find the directory on the filesystem.

dmgav:
[...]

PR #119, opened 2021-01-04
Loading of startup files based on IPython profile name.

dmgav:
The PR partially addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/114:

- Additional CLI option `--startup-profile` for `start-re-manager` is added, which allows to specify the name of IPython profile that contains the collection of startup files. The option works the same as `--startup-dir`, except that it does not require the user to specify path to the directory that contains startup files, but constructs the path based on the settings of IPython and provided profile name. The files are expected to be located at `<IPython_config_dir>/profile_<profile_name>/startup`. The parameter `--startup-profile` can not be used if IPython is not installed.

- The parameters `--startup-dir` and `--startup-profile` are mutually exclusive. The manager will fail to start if both parameters are specified.

- The existing parameter `--profile_collection` or `-p` is replaced by `--startup-dir`, since the old name does not accurately reflects the meaning.

- The parameter `--use_mpack` is renamed into `--use-persistent-metadata`, since the old name does not accurately describe the functionality.

- Removed restrictions on the names of startup files. In the original implementation, only files with names starting with two digits (`00` - `99`) were loaded. Now all files in the startup directory are loaded. The files are loaded in alphabetical order, so it i still recommended to number the files for convenience.

- Replaced `_` with `-` in all the other CLI parameters of `start-re-manager`:

### DIFF
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -1350,7 +1350,7 @@ def load_profile_collection_from_ipython(path=None):
         load_profile_collection_from_ipython()
     """
     ip = get_ipython()  # noqa F821
-    for f in sorted(glob.glob("[0-9][0-9]*.py")):
+    for f in sorted(glob.glob("*.py")):
         print(f"Executing '{f}' in TravisCI")
         ip.parent._exec_file(f)
     print("Profile collection was loaded successfully.")
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 qserver_version = bluesky_queueserver.__version__
 
 
-def get_default_profile_collection_dir():
+def get_default_startup_dir():
     """
     Returns the path to the default profile collection that is distributed with the package.
     The function does not guarantee that the directory exists.
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -205,8 +205,9 @@ def load_profile_collection(path, *, patch_profiles=True, keep_re=False):
     if not os.path.isdir(path):
         raise IOError(f"Failed to load the profile collection. Path '{path}' is not a directory.")
 
-    file_pattern = os.path.join(path, "[0-9][0-9]*.py")
-    file_list = glob.glob(file_pattern)
+    file_pattern_py = os.path.join(path, "*.py")
+    file_pattern_ipy = os.path.join(path, "*.ipy")
+    file_list = glob.glob(file_pattern_py) + glob.glob(file_pattern_ipy)
     file_list.sort()  # Sort in alphabetical order
 
     # If the profile collection contains no startup files, it is very likely
@@ -1350,7 +1351,7 @@ def load_profile_collection_from_ipython(path=None):
         load_profile_collection_from_ipython()
     """
     ip = get_ipython()  # noqa F821
-    for f in sorted(glob.glob("*.py")):
+    for f in sorted(glob.glob("*.py") + glob.glob("*.ipy")):
         print(f"Executing '{f}' in TravisCI")
         ip.parent._exec_file(f)
     print("Profile collection was loaded successfully.")
--- a/bluesky_queueserver/manager/start_manager.py
+++ b/bluesky_queueserver/manager/start_manager.py
@@ -274,10 +274,12 @@ def start_manager():
         profile_name = args.profile_name
         try:
             import IPython
+
             path_to_ipython = IPython.paths.get_ipython_dir()
         except Exception:
-            logger.error("IPython is not installed. Specify directory to startup file by using "
-                         "'--startup-dir' option.")
+            logger.error(
+                "IPython is not installed. Specify directory to startup file by using " "'--startup-dir' option."
+            )
             return 1
         startup_dir = os.path.abspath(path_to_ipython)
         profile_name_full = f"profile_{profile_name}"
--- a/bluesky_queueserver/manager/start_manager.py
+++ b/bluesky_queueserver/manager/start_manager.py
@@ -191,7 +191,7 @@ def start_manager():
         description="Start a RE Manager", epilog=f"blueksy-queueserver version {__version__}"
     )
     parser.add_argument(
-        "--zmq_addr",
+        "--zmq-addr",
         dest="zmq_addr",
         type=str,
         default="tcp://*:5555",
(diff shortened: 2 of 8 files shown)


bluesky_queueserver/manager/profile_ops.py
```diff
@@ -1350,7 +1350,7 @@ def load_profile_collection_from_ipython(path=None):
         load_profile_collection_from_ipython()
     """
     ip = get_ipython()  # noqa F821
-    for f in sorted(glob.glob("[0-9][0-9]*.py")):
+    for f in sorted(glob.glob("*.py")):
```
danielballan, line 1353: IPython also executes files named `*.ipy`. These can be used to invoke non-Python syntax, such as magics. This is the relevant method in IPython, for comparison. https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/core/shellapp.py#L403

bluesky_queueserver/manager/start_manager.py
```diff
@@ -188,24 +188,32 @@ def run(self):
...
-        "--profile_collection",
-        "-p",
-        dest="profile_collection_path",
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--startup-dir",
+        dest="startup_dir",
+        type=str,
+        help="Path to directory that contains startup files (profile collection).",
```
danielballan, line 206: I think `(profile collection)` might be confusing here because it might be not be obvious whether `~/.ipython/profile_collection` or `~/.ipython/profile_collection/startup` are expected. Is it too verbose to include `(example: ~/.ipython/profile_collection/startup)`

bluesky_queueserver/manager/start_manager.py
```diff
@@ -188,24 +188,32 @@ def run(self):
...
+    )
+    group.add_argument(
+        "--startup-profile",
+        dest="profile_name",
         type=str,
-        help="Path to directory that contains profile collection.",
+        help="The name of IPython profile used to find the location of startup files (profile collection).",
```
danielballan, line 212: Also here, `(profile collection)` makes it ambiguous whether `profile_collection` or just `collection` are expected. How about `(example: collection)`.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -214,36 +222,42 @@ def start_manager():
...
+        "--user-group-permissions",
         dest="user_group_permissions_path",
         type=str,
         help="Path to file that contains lists of plans and devices available to users. "
         "The path may be a relative path to the profile collection directory. "
         "If the path is a directory, then the default file name "
         "'user_group_permissions.yaml' is used.",
     )
-    parser.add_argument("--kafka_topic", type=str, help="The kafka topic to publish to.")
+    parser.add_argument("--kafka-topic", dest="kafka_topic", type=str, help="The kafka topic to publish to.")
```
danielballan, line 233: I believe that `dest` is unnecessary here and elsewhere. Argparse is smart enough to do the `-` -> `_` replacement by default.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -260,38 +274,54 @@ def start_manager():
...
-        pc_path = os.path.abspath(os.path.expanduser(pc_path))
-        if not os.path.exists(pc_path):
-            logger.error("Profile collection directory '%s' does not exist", pc_path)
-            return 1
-        if not os.path.isdir(pc_path):
-            logger.error("Path to profile collection '%s' is not a directory", pc_path)
+    # Find startup directory
+    if args.profile_name:
+        profile_name = args.profile_name
+        try:
```
danielballan, line 280: Prefer

```py
from importlib.util import find_spec

if find_spec("IPython"):
    import IPython
    path_to_ipython = ...
else:
    logger.error(...)
```

See [Slides 175-178](http://www.dabeaz.com/modulepackage/ModulePackage.pdf) for illustration of how `try`/`except` can be confusing here if you are curious.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -260,38 +274,54 @@ def start_manager():
...
+            import IPython
+
+            path_to_ipython = IPython.paths.get_ipython_dir()
+        except Exception:
+            logger.error(
+                "IPython is not installed. Specify directory to startup file by using " "'--startup-dir' option."
+            )
             return 1
+        startup_dir = os.path.abspath(path_to_ipython)
```
danielballan, line 289: This is redefined below to mean something else. Redefining local variables within the same scope and branch can make debugging confusing. Can we rename this variable `ipython_dir`?

bluesky_queueserver/manager/tests/_common.py
```diff
@@ -25,14 +25,14 @@ def copy_default_profile_collection(tmp_path, *, copy_yaml=True):
...
-    pc_path = get_default_profile_collection_dir()
+    pc_path = get_default_startup_dir()
     # New path
     new_pc_path = os.path.join(tmp_path, "startup")
 
     os.makedirs(new_pc_path, exist_ok=True)
 
     # Copy simulated profile collection (only .py files)
-    patterns = ["[0-9][0-9]*.py"]
+    patterns = ["*.py"]
```
danielballan, line 35: This will need to be updated to include `*.ipy` as well, as noted above.

danielballan (commented):
I like all of these changes. I have left some suggestions about the implementation.

bluesky_queueserver/manager/start_manager.py
dmgav: I added `dest` here for consistency with other arguments. It seems safer to define `dest`, because it decouples the parameter name (which could change in the future) and the variable name (which could remain the same even if parameter name is slightly modified).

bluesky_queueserver/manager/profile_ops.py
dmgav: Added `*.ipy` extension in all instances.

bluesky_queueserver/manager/start_manager.py
dmgav: `--startup-dir` may be any accessible directory, which contains meaningful `*.py` and `*.ipy` files. I changed the help string to reflect this point.

bluesky_queueserver/manager/start_manager.py
dmgav: I added example to the help string.

bluesky_queueserver/manager/start_manager.py
dmgav: Done.

bluesky_queueserver/manager/start_manager.py
dmgav: Done.

bluesky_queueserver/manager/tests/_common.py
dmgav: Changed in all instances.

dmgav:
@danielballan Thanks for detailed review. Changes are implemented.

danielballan (approved):
This looks great. The help strings in the CLI are very clear now.

PR #120, opened 2021-01-08
Populating worker namespace by loading a scripts or importing a module

dmgav:
The PR addresses the remaining requirements from issue https://github.com/bluesky/bluesky-queueserver/issues.
Changes:

- Additional parameters for `start-re-manager`: `--startup-module` followed by name of the module (on the module search path) and `--startup-script` followed by full path to the script to load.

- Scripts with local imports are not supported. The option to load scripts with local imports is implemented and tested, but disabled.

- CLI parameters for `qserver_list_of_plans_and_devices` are modified to support loading startup data from a script or a module.

- Unit tests covering new functionality are implemented.

### DIFF
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -379,6 +379,8 @@ def load_worker_startup_code(
 
     if startup_dir is not None:
         logger.info("Loading RE Worker startup code from directory '%s' ...", startup_dir)
+        startup_dir = os.path.abspath(os.path.expanduser(startup_dir))
+        print(f"startup_dir={startup_dir}") ##
         nspace = load_profile_collection(startup_dir, keep_re=keep_re)
 
     elif startup_module_name is not None:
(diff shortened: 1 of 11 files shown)


bluesky_queueserver/manager/profile_ops.py
```diff
@@ -250,6 +262,148 @@ def load_profile_collection(path, *, patch_profiles=True, keep_re=False):
...
+    _discard_re_from_nspace(nspace, keep_re=keep_re)
+
+    return nspace
+
+
+class StartupLoadingError(Exception):
+    ...
+
+
+def load_startup_script(script_path, *, keep_re=False, enable_local_imports=False):
```
danielballan, line 296: A flag that is off by default is a decent compromise here.

dmgav: Since I spent time experimenting with script loading, I decided to implement all the features that I could make work so that in the future we could easily turn them on if needed. The case of `enable_local_imports=True` is covered by unit tests and seem to work fine. 

I also would like to note, that RE Worker is loading a script/module/startup collection only once. In order to reload startup code one would close RE Worker (exit or terminate the process), then start a new process and then load startup code (again only once). So I expect that a lot of issues that could come up when a script or a module is loaded multiple times within the same process would not come up here.

danielballan (approved):
I approve this in principle. I have not had time for a detailed line-by-line review, but it can go in if it's blocking progress.

PR #122, opened 2021-01-13
Populating of worker namespace by devices from Happi

dmgav:
Changes:

- implemented a function `load_devices_from_happi` that populates the script namespace with devices using initialization data stored in Happi database. The function accepts a list of device names. The elements of the list could be strings or tuples containing two strings. If an element is a sting representing a device name, then the database is searched for the device with the name, the device is instantiated and loaded into the namespace. 
```
# In the startup script:
device_names = [
    'det',
    'motor1',
    'motor2',
]
load_devices_from_happi(device_names, namespace=locals())
# The devices `det`, `motor1` and `motor2` are loaded into the namespace
```
 
If an element is a tuple of two names, the first name is used to search the database for the device, the device name is changed to the second name in the tuple, then the device is instantiated and loaded into the database with the new name. The option that allows renaming the devices may be useful if devices from multiple beamlines are stored in the same database, and multiple beamlines decide to use (or are already using) devices with the identical convenient name in their startup code. For example if beamlines `ABC` and `DEF` use the device named `det`, the devices could be stored in common database as `abc_det` and `def_det`. The correct versions of the device may be loaded using the following startup code:

```
# In the startup script for beamline ABC:
device_names = [
    ('abc_det',  'det'),
]
load_devices_from_happi(device_names, namespace=locals())
# The correct version of the device is loaded into the namespace as 'det'
```

```
# In the startup script for beamline DEF
device_names = [
    ('def_det',  'det'),
]
load_devices_from_happi(device_names, namespace=locals())
# The correct version of the device is loaded into the namespace as 'det'
```

- The function is not using any features specific to QueueServer and may be used in any startup script, module or collection of startup files (profile collection).

- Detailed unit tests for `load_devices_from_happi()`.

- Implemented unit tests that demonstrate that the function `load_devices_from_happi` works with all supported methods for populating RE Worker namespace, including loading a script, importing a module or loading a profile collection.

### DIFF
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -149,7 +149,7 @@ def load_devices_from_happi(device_names, *, namespace=None, **kwargs):
         load_devices_from_happi(["det1", "motor1"], namespace=locals())
         # Works exactly the same as the previous example. If the second tuple element is
         #   evaluated as boolean value of False, then it is ignored and the device is not renamed.
-        load_devices_from_happi([("det1", None), ("motor1", "")], namespace=locals())
+        load_devices_from_happi([("det1", ""), ("motor1", "")], namespace=locals())
 
         # Load 'abc_det1' as 'det1' and 'abc_motor1' as 'motor1'.
         load_devices_from_happi([("abc_det1", "det1"), ("abc_motor1", "motor1")], namespace=locals())
@@ -161,7 +161,7 @@ def load_devices_from_happi(device_names, *, namespace=None, **kwargs):
         two elements. If an element is a tuple of two names, then the first name is used in database
         search and the second name is the name of the device after it is loaded into the namespace.
         Each element of the list is processed separately and may contain mix of strings and tuples.
-        If the second name in the tuple has boolean value of ``False``, then it is ignored and
+        If the second name in the tuple is an empty string, then it is ignored and
         the device is imported with the same name as used in database. It is expected that search
         will result in one found device per device name, otherwise ``RuntimeError`` is raised.
     namespace : dict
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -127,44 +127,48 @@ def set_user_ns(func):
     return wrapper
 
 
-def load_devices_from_happi(device_names, *, instrument, namespace=None, **kwargs):
+def load_devices_from_happi(device_names, *, namespace=None, **kwargs):
     """
-    Load the devices from Happi. The devices are specified as a list of device names.
-    Since the database may contain devices from multiple instruments (beamlines or endstations),
-    the function requires to specify the instrument name, that is included in each search request.
-    The instrument name could be arbitrary string, but it must be unique for the instrument
-    and consistently used throughout the database (``instrument`` metadata key). If instrument
-    name should not be included in the search parameters, then the function should be called with
-    ``beamline=None`` set explicitly. The ``instrument`` parameter is set as required to avoid
-    potential errors due to accidental loading devices configured for different instruments.
+    Load the devices from Happi based on the list of ``device_names``. The elements of the list
+    may be strings (device name) or tuples of two strings (device name used for database search and
+    the name with which the device is loaded in the script namespace). Two names may be needed
+    in case the devices are stored in the database using compound names that contain beamline
+    acronyms (e.g. ``abc_det`` and ``def_det`` for beamlines ABC and DEF) but they are expected
+    to have the name ``det`` when loaded during ABC and DEF beamline startup respectively (see examples).
 
-    The function may be called multiple times in a row if needed. For example, if the instrument
-    is split into two subsystems ``TEST.1`` and ``TEST.2``, then the devices from both subsystems
-    may be loaded into a script as
+    The devices are loaded into a namespace referenced by ``namespaced`` parameter.
+    The function may be called multiple times in a row for the same namespace to populate
+    it with results of multiple searches. The function also returns the dictionary of loaded devices,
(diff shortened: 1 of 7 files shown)


danielballan:
As an alternative to `(“name”, “custom name”)`, have you considered `(“canonical name”, {“name”: “custom name”})` or `{“canonical name”: {“name”: “custom name”}}`? These leave room for the possibility of overriding more than the name in the future. I like the latter for type stability reasons: both the simple case (list of canonical names) and the customized case (dict of canonical names mapped to custom options) place the canonical names as the thing that iteration would return. I prefer this to a heterogenous list of strings and tuples, which is harder to program against.

dmgav:
@danielballan I am not sure we would need to override anything by the device name. My initial idea was to use an additional field (say `instrument`) to limit search to devices that belong to a given beamline, and it worked fine if the database contained the respective entries (e.g. `{'name': 'det', 'instrument': 'SRX'}` and `{'name': 'det', 'instrument': 'HXN'}`). Unfortunately after I implemented this idea, I found out that the only way to achieve it is to manually edit the database (JSON file in this case). Happi is always setting `_id` of the database document equal to `name` before the document is saved. This is by design and I see reasoning for this. But the consequence is that names in the database must be unique. This prevents multiple beamlines from using devices that have the same name, but different set of parameters. So the solution is to prefix the device name with beamline/endstation/instrument acronym. This is how Happi is used at LCLS. If beamline staff agrees to change names of their devices to prefixed form, then the problem is solved, but @tacaswell mentioned that this is unlikely to happen. So the solution is to provide the ability to rename the devices in the startup script, and renaming should happen before devices are instantiated so that they are properly initialized. I think that if any other fields need to be changed, it has to be done in the database.

dmgav:
On the other hand we may want to change `prefix` to switch from hardware to simulated devices. So in principle we could implement the third option where entries of the list are dictionary:
```
[
    ("canonical_name", {"name": "custom_name", "prefix": "DDD"}),
    ("canonical_name", {"prefix": "DDD"}),  # In this case we don't necessarily need to change the name.
    ...
]

**Comment by danielballan** [comment]
I think three options is a lot of options. How about we support just the most explicit/flexible one at first and then add shortcuts as we learn more about how we use it?

**Comment by dmgav** [comment]
I was thinking about the workflow. Actually everything except Option 1 is a hack. Option 2 is hard coded and seems safe. Option 3 would expose the internals of parameter presentation to users and probably should not be implemented at all. I think that practically we will use Option 1 if we decide to have one database per beamline/endstation and Option 2 if we decide to use common database. I also think that the database should contain correct values of parameters. If a parameter needs to be changed, it should be changed in the database. Having one set of parameters in the database and then changing a lot of them in the startup script defeats the purpose of the database: we can just instantiate all the devices in the script. The only reason we need to change `name` is because it is used as document `_id` in the database and has to be unique.

As for simulation, I think that the test suite should contain a script that will load devices from the database, change their prefixes (if needed) and then saves them to a separate local database (JSON file). The test environment should be configured to use the local JSON-based database, so the startup script will not have to be modified. The JSON file can also be used to run tests on CI.

**Comment by tacaswell** [comment]
I think the only option we should put in is the minimal escape hatches we need.  The mid-to-long term vision is that this data base is co-maintained by the people who maintain the IOCs and should be a derived product from the canonical information about what devices / IOCs are available and configured.   

We still have the escape hatch of "just instantiate the device in Python".

I think the ability to replace just the name is also consistent with the rest of our designs as this is what the Components are doing (giving a locally unique but readable name for a globally unique but annoyingly verbose name).

File: bluesky_queueserver/manager/profile_tools.py [hunk 1]
```diff
@@ -125,3 +126,154 @@ def wrapper(*args, **kwargs):
...
+def load_devices_from_happi(device_names, *, namespace=None, **kwargs):
+    """
+    Load the devices from Happi based on the list of ``device_names``. The elements of the list
+    may be strings (device name) or tuples of two strings (device name used for database search and
+    the name with which the device is loaded in the script namespace). Two names may be needed
+    in case the devices are stored in the database using compound names that contain beamline
+    acronyms (e.g. ``abc_det`` and ``def_det`` for beamlines ABC and DEF) but they are expected
+    to have the name ``det`` when loaded during ABC and DEF beamline startup respectively (see examples).
+
+    The devices are loaded into a namespace referenced by ``namespaced`` parameter.
```
tacaswell, line 140:
```suggestion
    The devices are loaded into a namespace referenced by ``namespace`` parameter.
```

tacaswell:
I have a slight concern about the mutate-in-place namespace vs return namespace.  I think it should either take a required namespace to fill and return a list of the items added or return a dict of the name / devices and make the caller update whatever namespace they want.

dmgav:
@tacaswell Those are two separate use cases for the function. I think that the most useful is the one that updates the namespace passed as a reference: this is how it is supposed to be used in startup scripts. The second option is to take the returned dictionary (identical dictionary may be obtained by passing an empty dictionary as a `namespace` parameter) and do arbitrary operations with it. So the second option is redundant. I guess from software engineering perspective it makes sense to get rid of redundancy. Returning the list may be a good idea.

bluesky_queueserver/manager/profile_tools.py
```diff
@@ -274,6 +287,5 @@ def load_devices_from_happi(device_names, *, namespace=None, **kwargs):
 
     ns = load_devices(*[_.item for _ in results])
     ns_dict = ns.__dict__
-    if namespace:
-        namespace.update(ns_dict)
-    return ns_dict
+    namespace.update(ns_dict)
+    return list(ns_dict.keys())
```
tacaswell, line 291:
```suggestion
    return list(ns_dict)
```

tacaswell (approved):
One minor code nit.

Anyone can merge with or without applying the suggestion.

PR #124, opened 2021-02-10
Subscribing RE to 0MQ Publisher

dmgav:
A built-in option to subscribe Run Engine to `bluesky.callbacks.zmq.Publisher` is added. The address of 0MQ proxy is passed with `start-re-manager` CLI parameter `--zmq-data-proxy-addr`. Identically to subscription to Kafka, the option works for Run Engine created by RE Manager and in the startup script (`start-re-manager` called with `--keep-re`). If the parameter `--zmq-data-proxy-addr` is not used, RE is not subscribed to the Publisher callback.

### DIFF
--- a/bluesky_queueserver/manager/start_manager.py
+++ b/bluesky_queueserver/manager/start_manager.py
@@ -272,7 +272,7 @@ def start_manager():
         type=str,
         help="The address of ZMQ proxy used to publish data. If the parameter is specified, RE is "
         "be subscribed to 'bluesky.callbacks.zmq.Publisher' and documents are published via 0MQ proxy. "
-        "0MQ Proxy (see Bluesky 0MQ documentation) should be started before plans could be executed. "
+        "0MQ Proxy (see Bluesky 0MQ documentation) should be started before plans are executed. "
         "The address should be in the form '127.0.0.1:5567' or 'localhost:5567'. The address is passed "
         "to 'bluesky.callbacks.zmq.Publisher'. It is recommended to use Kafka instead of 0MQ proxy in "
         "production data acquisition systems and use Kafka instead.",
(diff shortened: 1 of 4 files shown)


bluesky_queueserver/manager/start_manager.py
```diff
@@ -264,6 +265,19 @@ def start_manager():
         help="Bootstrap server to connect to.",
         default="127.0.0.1:9092",
     )
+
+    parser.add_argument(
+        "--zmq-data-proxy-addr",
+        dest="zmq_data_proxy_addr",
+        type=str,
+        help="The address of ZMQ proxy used to publish data. If the parameter is specified, RE is "
+        "be subscribed to 'bluesky.callbacks.zmq.Publisher' and documents are published via 0MQ proxy. "
```
danielballan, line 274: The should say "RE is to be" or "RE is" or "RE will be". Not sure which one you meant.

bluesky_queueserver/manager/start_manager.py
```diff
@@ -264,6 +265,19 @@ def start_manager():
...
+        "0MQ Proxy (see Bluesky 0MQ documentation) should be started before plans could be executed. "
```
danielballan, line 275:
```suggestion
        "0MQ Proxy (see Bluesky 0MQ documentation) should be started before plans are executed. "
```

danielballan (approved):
I have some wording suggestions to consider. Otherwise, please merge at will.

bluesky_queueserver/manager/start_manager.py
dmgav: I meant 'is'.

dmgav: Fixed.

mrakitin (commented):
Late to the party, but looks good to me!

PR #129, opened 2021-03-02
API for detecting if the script is executed in RE Worker environment.

dmgav:
Implementation of API function `is_re_worker_active()` for using in startup scripts or modules. The API function returns boolean value that tells if the script is executed in RE Worker environment. One potential use of the API is to disable interactive features in a script (such as manual input) when the script is executed autonomously in RE Worker environment:

```
from bluesky_queueserver.manager.profile_tools import is_re_worker_active

...

if is_re_worker_active():
    (code without interactive features, e.g. reading data from a file)
else:
    (code with interactive features, e.g. manual data input)

...
```
`is_re_worker_active()` may be used in any code executed by the RE Worker process including the modules imported by startup scripts. 

The `is_re_worker_active()` function checking if an environment variable `QSERVER_RE_WORKER_ACTIVE` is set in the current process and returns `True` if the variable is found and `False` otherwise. The environment variable is set/unset by the additional API functions `set_re_worker_active()` and `clear_re_worker_active()`. The functions add/remove elements in the dictionary `os.environ` and therefore influence only the process in which they are called.

The additional API functions `set_re_worker_active()` and `clear_re_worker_active()` are not designed to be used in startup scripts, but could be used to emulate RE Worker environment and force execution of the scripts in autonomous mode (see implementation of the function ``gen_list_of_plans_and_devices`` in ``bluesky_queueserver.manager.profile_ops`` module).

Addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/133

### DIFF
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -301,7 +301,7 @@ def set_re_worker_active():
     Subsequent calls to ``is_re_worker_active`` in the current process will return ``True``.
     This function should not be used in the startup scripts.
     """
-    os.environ[_env_re_worker_active] = True
+    os.environ[_env_re_worker_active] = ""
 
 
 def clear_re_worker_active():
@@ -338,4 +338,4 @@ def is_re_worker_active():
     boolean
         ``True`` - the code is executed in RE Worker environment, otherwise ``False``.
     """
-    return bool(os.environ.get(_env_re_worker_active, False))
+    return _env_re_worker_active in os.environ
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -301,7 +301,7 @@ def set_re_worker_active():
     Subsequent calls to ``is_re_worker_active`` in the current process will return ``True``.
     This function should not be used in the startup scripts.
     """
-    os.environ[_env_re_worker_active] = ""
+    os.environ[_env_re_worker_active] = "1"
 
 
 def clear_re_worker_active():
@@ -338,4 +338,4 @@ def is_re_worker_active():
     boolean
         ``True`` - the code is executed in RE Worker environment, otherwise ``False``.
     """
-    return _env_re_worker_active in os.environ
+    return bool(os.environ.get(_env_re_worker_active, False))
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -338,4 +338,4 @@ def is_re_worker_active():
     boolean
         ``True`` - the code is executed in RE Worker environment, otherwise ``False``.
     """
-    return bool(os.environ.get(_env_re_worker_active, False))
+    return os.environ.get(_env_re_worker_active, "false").lower() not in ("false", "0", "")
--- a/bluesky_queueserver/manager/profile_tools.py
+++ b/bluesky_queueserver/manager/profile_tools.py
@@ -1,6 +1,7 @@
 import functools
 import inspect
 import re
+import os
 
 
 class UserNamespace:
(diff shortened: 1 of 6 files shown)


bluesky_queueserver/manager/profile_tools.py
```diff
@@ -289,3 +290,52 @@ def load_devices_from_happi(device_names, *, namespace, **kwargs):
...
+_env_re_worker_active = "QSERVER_RE_WORKER_ACTIVE"
+
+
+def set_re_worker_active():
+    """
+    Set the environment variable used to determine if the current process is RE Worker process.
+    Subsequent calls to ``is_re_worker_active`` in the current process will return ``True``.
+    This function should not be used in the startup scripts.
+    """
+    os.environ[_env_re_worker_active] = ""
```
mrakitin, line 304: Should the value be "True" or "1" instead of an empty string? We may want to do something like the following to reset it:
```bash
$ QSERVER_RE_WORKER_ACTIVE= <our command>
```
and this will enable it:
```bash
$ QSERVER_RE_WORKER_ACTIVE=1 <our command>
```

bluesky_queueserver/manager/profile_tools.py
```diff
@@ -289,3 +290,52 @@ def load_devices_from_happi(device_names, *, namespace, **kwargs):
...
+            (code with interactive features, e.g. manual data input)
+
+        ...
+
+    Returns
+    -------
+    boolean
+        ``True`` - the code is executed in RE Worker environment, otherwise ``False``.
+    """
+    return _env_re_worker_active in os.environ
```
mrakitin, line 341: This condition will need to be updated if the previous suggestion is implemented.

mrakitin (approved):
Looks good. I have an optional suggestion to assign a non-empty value to the environment variable (see below), rather than just testing for the env variable presence.

bluesky_queueserver/manager/profile_tools.py
dmgav: I don't think that we are going to be setting this variable externally, but it can be changed I guess.

mrakitin: I see. As it's going to be used from inside the code, maybe it's worth adding a leading underscore (such as `_QSERVER_RE_WORKER_ACTIVE`) to mark it as a "private" env var (strictly speaking, bash does not care about it, but we can follow Python's style here).

dmgav: I am not sure about changing the environment variable name. May be it is better to keep it standard.

mrakitin:
I just tried a simple implementation of the latest commit, and the second case does not look right:
```bash
$ QSERVER_RE_WORKER_ACTIVE=1 python -c "import os; print(bool(os.environ.get('QSERVER_RE_WORKER_ACTIVE', False)))"
True
$ QSERVER_RE_WORKER_ACTIVE=0 python -c "import os; print(bool(os.environ.get('QSERVER_RE_WORKER_ACTIVE', False)))"
True
$ QSERVER_RE_WORKER_ACTIVE= python -c "import os; print(bool(os.environ.get('QSERVER_RE_WORKER_ACTIVE', False)))"
False
$ python -c "import os; print(bool(os.environ.get('QSERVER_RE_WORKER_ACTIVE', False)))"
False
```

bluesky_queueserver/manager/profile_tools.py
dmgav: In principle I wouldn't exclude potential to use the command like this: `$ QSERVER_RE_WORKER_ACTIVE=1 <our command>`.

dmgav:
I think this is because "0" is a string.

dmgav:
So it differentiates between empty string / non-empty string

mrakitin:
> So it differentiates between empty string / non-empty string.

Yes, bash does not have types, Python recognizes them with types, so we need a couple of more lines to check types/convert properly.

dmgav:
I don't see Python recognizing types unless I try to explicitly convert variable to int.

mrakitin:
I meant it's also a string in Python, so yes, we need conversion and a separate check for an empty string before conversion.

mrakitin (approved):
I like the version without `bool`. A slight improvement can be to use `os.getenv(..., 'false')`, but it's not critical.

dmgav:
I think that using `os.environ` in all API is more consistent.

mrakitin:
> I think that using `os.environ` in all API is more consistent.

Agree, as long as `.get()` with the default value is added, that should be safe enough.

Regarding the env var setting from the outside, I looked at another way to accessing 0/1 as bools using `distutils.util.strtobool(some_string)` (https://stackoverflow.com/a/18472142). It seems to be a nice alternative, while the implementation is very similar (after adding a `bool(...)` wrapper):

```py
In [22]: distutils.util.strtobool??
Signature: distutils.util.strtobool(val)
Source:
def strtobool (val):
    """Convert a string representation of truth to true (1) or false (0).

    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
    'val' is anything else.
    """
    val = val.lower()
    if val in ('y', 'yes', 't', 'true', 'on', '1'):
        return 1
    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
        return 0
    else:
        raise ValueError("invalid truth value %r" % (val,))
```
People also use `json` for conversion, but it may be too much for that simple task.

dmgav:
I think that it is very unlikely that the variable will ever be set externally. I see no use case for this except some rare debugging cases. There is no sense in running startup script from command line, so there will always be a code wrapper for this. There are 'official' API: `set_re_worker_active` and `clear_re_worker_active` for use in Python code.  I will extend the list of values indicating `False` to `('n', 'no', 'f', 'false', 'off', '0', '')`, all other values will be considered as `True`.

Issue #130, opened 2021-03-03
Server logs (https server and manager) need to be quieter


mrakitin:
Task for 5L demo.

Server logs (https server and manager) need to be quieter. Sometimes client times out because the web server is busy printing 10k lines of JSON.

dmgav:
Resolved in PR https://github.com/bluesky/bluesky-queueserver/pull/135

PR #135, opened 2021-03-10
CLI arguments for 'start-re-manager' to set logging verbosity

dmgav:
The set of CLI parameters was added to `start-re-manager` that allow to modify verbosity of logging output. Currently logging can only be sent to stdout/stderr. By default, RE Manager will display log messages starting from `loglevel=INFO`. The log includes warnings, errors, brief messages with information about the state of the manager and worker processes and executed steps. This behavior may be modified by calling `start-re-manager` with the following mutually exclusive parameters:

- `--verbose` prints all logging messages including `DEBUG` messages. In addition to short messages indicating the executed steps, the output will display contents of all received/send messages, submitted plans etc. Verbose output may be useful for debugging purposes, but may slow down the system if the plan queue, history or the lists of plans/devices become long.

- `--quiet` prints only warning and error message. Output is insufficient to monitor plan execution.

- `--silent` disables logging. 

Resolves the issue https://github.com/bluesky/bluesky-queueserver/issues/130

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -896,7 +896,7 @@ class RunEngineManager(Process):
                 break
         if item_type is None:
             raise Exception(
-                "Incorrect request format: request contains no item info. Supported item types: {item_types}"
+                f"Incorrect request format: request contains no item info. Supported item types: {item_types}"
             )
         return item, item_type
 
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -196,17 +196,17 @@ class RunEngineManager(Process):
             success = await self._watchdog_start_re_worker()
             if not success:
                 raise RuntimeError("Failed to create Worker process")
-            logger.debug("Waiting for RE worker to start ...")
+            logger.info("Waiting for RE worker to start ...")
             success = await self._fut_manager_task_completed  # TODO: timeout may be needed here
             if success:
                 self._environment_exists = True
-                logger.debug("Worker started successfully.")
+                logger.info("Worker started successfully.")
                 success, err_msg = True, ""
             else:
                 self._environment_exists = True
                 await self._confirm_re_worker_exit()
                 self._environment_exists = False
-                logger.debug("Error occurred while opening RE Worker environment.")
+                logger.error("Error occurred while opening RE Worker environment.")
                 success, err_msg = False, "Error occurred while opening RE Worker environment."
 
         except Exception as ex:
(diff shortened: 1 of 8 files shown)


bluesky_queueserver/manager/manager.py
```diff
@@ -1521,8 +1522,7 @@ def run(self):
         Overrides the `run()` function of the `multiprocessing.Process` class. Called
         by the `start` method.
         """
-
-        logging.basicConfig(level=logging.WARNING)
+        logging.basicConfig(level=max(logging.WARNING, self._log_level))
```
mrakitin, line 1525: What is the intention here? E.g. if `self._log_level` is `INFO` (=20), the set level will be `WARNING` (=30). The description says the default is `INFO`.

bluesky_queueserver/manager/plan_monitoring.py
```diff
@@ -120,8 +120,8 @@ def start(self, doc):
             uid = doc["uid"]
             self._run_list.add_run(uid=uid)
 
-            print(f"New run was open: '{uid}'")
-            print(f"Run list: {self._run_list.get_run_list()}")
+            logger.info(f"New run was open: '{uid}'")
+            logger.debug(f"Run list: {self._run_list.get_run_list()}")
```
mrakitin, line 124: I saw you used the `%s` style for logging above, which is considered more performant. Do you want to do the same here?

bluesky_queueserver/manager/start_manager.py
```diff
@@ -312,10 +312,46 @@ def start_manager():
         help="Name of the Data Broker configuration file.",
     )
 
-    logging.basicConfig(level=logging.WARNING)
-    logging.getLogger("bluesky_queueserver").setLevel("DEBUG")
+    group_verbosity = parser.add_argument_group(
+        "Logging verbosity settings",
+        "The default logging settings (loglevel=INFO) provide optimal amount of data to monitor "
+        "the operation of RE Manager. Select '--verbose' option to see detailed data on received and "
+        "sent messages, addeded and executed plans etc. Use options '--quiet' and '--silent' to "
```
mrakitin, line 319:
```suggestion
        "sent messages, added and executed plans, etc. Use options '--quiet' and '--silent' to "
```

mrakitin (commented):
I am only concerned about `logging.basicConfig(level=max(logging.WARNING, self._log_level))` (left a comment), otherwise looks good.

bluesky_queueserver/manager/manager.py
dmgav: This would be logging level for other packages except `bluesky_queueserver` for which logging is not set explicitly. Typically it is undesirable to get debug/info logging from all packages in any circumstances. So here I limit logging level to WARNING or higher.

bluesky_queueserver/manager/plan_monitoring.py
dmgav: Yes

bluesky_queueserver/manager/manager.py
```diff
@@ -896,7 +896,7 @@ def _get_item_from_request(self, *, request):
                 break
         if item_type is None:
             raise Exception(
-                "Incorrect request format: request contains no item info. " f"Supported item types: {item_types}"
+                "Incorrect request format: request contains no item info. Supported item types: {item_types}"
```
mrakitin, line 899:
```suggestion
                f"Incorrect request format: request contains no item info. Supported item types: {item_types}"
```

mrakitin (approved):
Looks good to me now. Thanks for accepting the suggestions, @dmgav!

Issue #141, opened 2021-03-23
More detailed log messages for RE Manager running in normal mode 


dmgav:
The changes in PR https://github.com/bluesky/bluesky-queueserver/pull/135 included options to change logging verbosity for RE Manager. In normal mode (started without explicitly specifying verbosity level), RE Manager displays messages with level `INFO` and above. Some of the messages are not sufficiently informative to monitor operation of RE Manager. For example, when a new plan is started, the log should contain some information about the plan (plan name, parameters, user name etc.) without overloading logging input with debug level information. Logging messages should be reviewed and information should be added wherever necessary.

dmgav:
Addressed in PR https://github.com/bluesky/bluesky-queueserver/pull/142

PR #140, opened 2021-03-22
0MQ API for adding a batch of items to the back of the queue ('queue_item_add_batch')

dmgav:
Implementation of `queue_item_add_batch` 0MQ API. Each item in the batch is validated and the the batch is added to the queue only if each item successfully passes validation. The returned item list contains `success` status for and error message for each item. The returned information is sufficient for generating detailed reports that could be displayed to the user.

The handler for `/queue/upload/spreadsheet` is modified to use `queue_item_add_batch`. The return parameter `result` was renamed to `item_list` (I don't think the parameter was used in any code yet).

The API documentation is added to 0MQ API section.

The PR addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/139

PR #142, opened 2021-03-24
More detailed INFO log messages

dmgav:
Added more detailed INFO log message for operations of starting a plan (prints plan parameters), executing queue instruction (prints instruction parameters). Also added brief INFO messages for indicating the results of execution of `queue_item_add`, `queue_item_add_batch` and `queue_item_update` API requests.

Addresses issue https://github.com/bluesky/bluesky-queueserver/issues/141

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -1003,6 +1003,22 @@ class RunEngineManager(Process):
 
         return item, item_uid_original
 
+    def _generate_item_log_msg(self, prefix_msg, success, item_type, item, qsize):
+        """
+        Generate short log message for reporting results of ``queue_item_add``, ``queue_item_add_batch``
+        and ``queue_item_update`` operations.
+        """
+        log_msg = f"{prefix_msg}: success={success} item_type='{item_type}'"
+        if item:
+            if "name" in item:
+                log_msg += f" name='{item['name']}'"
+            elif "action" in item:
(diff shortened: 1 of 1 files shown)


bluesky_queueserver/manager/manager.py
```diff
@@ -1032,12 +1066,114 @@ async def _queue_item_add_handler(self, request):
...
+        except Exception as ex:
+            success = False
+            msg = f"Failed to add an item: {str(ex)}"
+
+        try:
+            qsize = await self._plan_queue.get_queue_size()
+        except Exception:
+            qsize = None
+
+        logger.info(self._generate_item_log_msg("Batch of items added", success, None, None, qsize))
```
mrakitin, line 1170: Wouldn't it be better to use `batch` instead of the first `None`?

mrakitin (approved):
I assume it contains commits from #140, so I reviewed in this mode: https://github.com/dmgav/bluesky-queueserver/compare/zmq-items-batch...dmgav:log-details.

I have a single comment/suggestion, but it's not critical.

bluesky_queueserver/manager/manager.py
dmgav: I am not sure. `item_type` should be `plan` or `instruction`. For the batch, `item_type` is set for each item in the list. In the log this statement will be preceded with `"Adding a batch of items to the queue ..."`. The statement itself mentions that the batch is added.

Issue #138, opened 2021-03-22
Default (built-in) implementation for the function 'spreadsheet_to_plan_list' (HTTP Server)


dmgav:
PR https://github.com/bluesky/bluesky-queueserver/pull/137 implements REST API that accepts Excel spreadsheet, converts the spreadsheet to a list of plans and add the plans to the queue. In the existing implementation of API is using the function `spreadsheet_to_plan_list` provided in external beamline-specific module. The code also contains placeholder for the built-in function that is supposed to be used as default implementation in cases when external module is not loaded or the spreadsheet is rejected by the beamline-specific function (the external function returns `None`).

Implement the default processing function `spreadsheet_to_plan_list` that accepts Excel/Open Office (.xlsx, .xls, .odt etc.) single-sheet workbooks or .csv files. Spreadsheet format: top row contains parameter names, the remaining rows contain plan parameters. Columns: first column - plan name, second column - args (comma separated as in the list), the remaining columns are kwargs values (kwargs keys are column names in the first rows).

dmgav:
Addressed in PR https://github.com/bluesky/bluesky-queueserver/pull/144

PR #144, opened 2021-03-25
Implementation for a default spreadsheet processing function

dmgav:
Implementation for default (built-in) function for converting a spreadsheet to a list of plans. The plan parameters are expected to be represented as a table in straightforward way following guidelines. Users that require support for more sophisticated or custom spreadsheets may write their own processing function. The function supports `.xlsx` (Excel) and `.csv` files. Only the first sheet of an Excel workbook will be loaded.

The screenshot of an Excel spreadsheet with sample simulated plans:

Spreadsheet with sample plans 2021-03-25 13-15-19

```
    Guidelines:

    - The first row contains column names. The names of columns 1 and 2 can be
      arbitrary, but should be kept informative. It is recommended to name column 2 as
      'args'. The remaining column names should strictly match names of plan kwargs.

    - Each of the remaining rows contains parameters of a single plan. Empty rows
      are permitted. The row is considered empty if the cell containing plan name
      is empty. In this case the remaining row cells are ignored.

    - Column 1 contains plan names. Plan names should be enclosed in quotation marks.
      (Plan name will be accepted without quotation marks but it is recommended to use
      quatation marks for consistency.) Plan names must satisfy requirements for Python
      identifiers, otherwise the exception will be raised. (The plan with the respective
      name should also exist in the RE Worker environment, but this function doesn't
      verify it.)

    - Column 2 contains plan args. The args should be comma-separated. The list of args
      should not be enclosed in ``[]`` (e.g. ['det1', 'det2'] is considered as one arg
      containing a list of detectors. The names of the devices (motors or detectors)
      should be enclosed in quotation marks. The cell may be left empty if the plan
      is not using args. NOTE: column 2 is reserved for args and it should be present
      even if no plans in the list accepts args.

    - Columns 3,4... contain values of kwargs. Each column is reserved for kwarg with
      the name listed in row 1. If a kwarg is not used for a given plan, the cell should
      be empty.

    The contents of the cells are evaluated using Python expression ``eval(<cell_value>, {}, {}``,
    therefore the expressions may not contain variable names or function calls. But they may
    represent arbitrary expressions containing numbers and strings (numbers and strings
    can be organized in lists, dictionaries, lists of dictionaries etc. If a cell contains
    unquoted device name (e.g. ``motor``), the name will be interpreted by ``eval()`` as
    a variable and an exception will be raised.

```

Example of `.xlsx` file:
[spreadsheet.xlsx](https://github.com/bluesky/bluesky-queueserver/files/6206636/spreadsheet.xlsx)

Example of `.csv` file (generated by Pandas from a DataFrame with plan parameters):
```
plan_name,args,num,delay,detectors
'count',"['det1', 'det2']",10.0,,
'count',['det1'],,0.5,
'count',,,0.5,['det1']
,,,,
'scan',"['det1', 'det2'], 'motor', -1, 1",10.0,,
'scan',"['det1', 'det2'], 'motor', -1, 1, 10",,,
'count',['det2'],2.0,0.7,
'count',['det2'],2.0,,
```

Addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/138

### DIFF
--- a/bluesky_queueserver/server/conversions.py
+++ b/bluesky_queueserver/server/conversions.py
@@ -219,6 +219,10 @@ def spreadsheet_to_plan_list(*, spreadsheet_file, file_name, **kwargs):  # noqa:
     # The number of rows (rows are potentially plans, there could be empty rows though)
     n_rows = len(df[key_plan_name])
 
+    def _format_row(df, nr):
+        # Print the row as a tuple
+        return str(tuple([df[_][nr] for _ in df.keys()]))
+
     plan_list = []
 
     for nr in range(n_rows):
@@ -228,14 +232,24 @@ def spreadsheet_to_plan_list(*, spreadsheet_file, file_name, **kwargs):  # noqa:
         if not plan_name or _is_nan(plan_name):
             continue
 
+        # The following step allows the plan names to be quoted or unquoted as well as to detect cases
+        #   when name is of wrong type (list or dict) because spreadsheet formatting is incorrect.
+        try:
+            plan_name = _read_cell_parameter(plan_name)
+        except Exception:
+            # Exception is raised if the plan name is unquoted. If plan name is just a corrupt
+            #   expression, then it will remain a string and the plan will be rejected by RE Manager.
+            pass
         # Check if the plan name is string and it is formatted to represent a valid plan name
         if not isinstance(plan_name, str):
             raise ValueError(
-                f"Plan name '{plan_name}' (row {nr + 1}) is of incorrect type ('{type(plan_name)}'): "
-                "supported type 'str')"
+                f"Plan name '{plan_name}' (row {nr + 1}: {_format_row(df, nr)}) has "
+                f"incorrect type ('{type(plan_name)}'): only 'str' type is supported)"
             )
         if not plan_name.isidentifier():
-            raise ValueError(f"Plan name '{plan_name}' (row {nr + 1}) is not a valid plan name")
+            raise ValueError(
+                f"Plan name '{plan_name}' (row {nr + 1}: {_format_row(df, nr)}) is not a valid plan name"
+            )
 
         try:
             plan_args = []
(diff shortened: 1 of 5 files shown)


gwbischof:
Why should plan names have quotes?

dmgav:
@gwbischof Plan names can be without quotes, but I would use quotes. The rest of the names should be quotes. But if you are working on BMM GUI, this PR and the spreadsheet format is irrelevant. See https://github.com/NSLS-II-BMM/bluesky-httpserver-bmm Also the description in https://github.com/NSLS-II-BMM/bluesky-httpserver-bmm/pull/1

mrakitin (commented):
Why not to use the `dtype` kwarg from https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html to assign the data types for specific columns?

dmgav:
@mrakitin But how do you know the data type for a column? Also, a column may (and often will) contain elements of different types, so types of column elements almost always 'object'. Actually, `pandas` poorly fits this task. The only advantage of `pandas` is that it can read multiple file formats (like .csv) using almost the same (but not always the same) code. For processing of custom spreadsheets one should use `openpyxl` directly.

mrakitin:
Is there any reason to have heterogeneous data types in a single column? If so, I agree with your approach.

dmgav:
The reason is that the columns contain user defined values of `args` and `kwargs` of a plan. In Python developers and users are encouraged to use the same parameter to pass values arbitrary types then decide how to process those values depending on their type. So I would assume that the columns will very often have heterogeneous types. Actually I thing the only downside of using heterogeneous types in Pandas is computational efficiency, but those are going to be very small tables.

mrakitin (commented):
I was looking through the code before this PR was merged, and did not have any suggestions. Nice addition, @dmgav!

dmgav:
@mrakitin Thanks. Also I experimented with `dtype` parameter. It doesn't do what you would expect it to do. Instead it puts restrictions on types that could be present in the columns: if data type doesn't match it raises the exception. Also `dtype=str` forces column types to `object` and doesn't change anything. Assuming that we want to read spreadsheets provided by the user, we should assume that cell data may be of any type and then try to make sense of the data if possible with minimum restrictions.

Issue #148, opened 2021-04-07
Option to pass the address of RE Manager to HTTP Server


dmgav:
Implement an option to pass the ZMQ address for control communication channel to HTTP Server. Currently HTTP Server is always using the default ZMQ server address. The arbitrary address could be passed to the server by setting environment variable.

dmgav:
Addressed in PR https://github.com/bluesky/bluesky-queueserver/pull/149

PR #149, opened 2021-04-07
Option to set custom 0MQ address for communication between HTTP Server and RE Manager

dmgav:
Implemented an option to pass custom RE Manager 0MQ address to HTTP Server by setting the environment variable `QSERVER_ZMQ_ADDRESS`, e.g.

```
export QSERVER_ZMQ_ADDRESS='tcp://localhost:60615'
```
Unit tests for operation of RE Manager and HTTP Server using 0MQ address different from default.

Additional changes:

- Changed log level for message acknowledging request of RE Manager status to `debug` (to clear the log).

Addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/148

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,6 +92,12 @@ The Web Server should be started from the second shell as follows::
 
   uvicorn bluesky_queueserver.server.server:app --host localhost --port 60610
 
+The Web Server connects to RE Manager using Zero MQ. The default ZMQ address is 'tcp://localhost:60615'.
+A different ZMQ address may be passed to the Web Server by setting *QSERVER_ZMQ_ADDRESS* environment
+variable before starting the server::
+
+  export QSERVER_ZMQ_ADDRESS='tcp://localhost:60615'
+
 The Web Server supports using external modules for processing some requests. Those modules
 are optional and may contain custom instrument-specific processing code. The name of the external
 module may be passed to HTTP server by setting **BLUESKY_HTTPSERVER_CUSTOM_MODULE** environment
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,7 @@ The Web Server should be started from the second shell as follows::
   uvicorn bluesky_queueserver.server.server:app --host localhost --port 60610
 
 The Web Server connects to RE Manager using Zero MQ. The default ZMQ address is 'tcp://localhost:60615'.
-A different ZMQ address may be passed to the Web Server by setting *QSERVER_ZMQ_ADDRESS* environment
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -92,6 +92,12 @@ The Web Server should be started from the second shell as follows::
 
   uvicorn bluesky_queueserver.server.server:app --host localhost --port 60610
 
+The Web Server connects to RE Manager using Zero MQ. The default ZMQ address is 'tcp://localhost:60615'.
+A different ZMQ address may be passed to the Web Server by setting *QSERVER_ZMQ_ADDRESS* environment
```
mrakitin, line 96:
```suggestion
A different ZMQ address may be passed to the Web Server by setting the *QSERVER_ZMQ_ADDRESS* environment
```

bluesky_queueserver/manager/start_manager.py
```diff
@@ -207,7 +207,7 @@ def start_manager():
         dest="zmq_addr",
         type=str,
         default="tcp://*:60615",
-        help="The address of ZMQ server (control connection).",
+        help="The address of ZMQ server (control connection), e.g. 'tcp://*:60615.'",
```
mrakitin, line 210:
```suggestion
        help="The address of ZMQ server (control connection), e.g. 'tcp://*:60615'.",
```

bluesky_queueserver/server/server.py
```diff
@@ -44,8 +44,12 @@ async def startup_event():
 
     # TODO: implement nicer exit with error reporting in case of failure
 
+    zmq_server_address = os.getenv("QSERVER_ZMQ_ADDRESS", None)
```
mrakitin, line 47: I was thinking about having those common variables (such as the one here `QSERVER_ZMQ_ADDRESS`, and the others related to encryption) defined in a module somewhere, and they can be imported and reused from a single location, in a similar fashion as it was recently done in https://github.com/caproto/caproto/pull/727. That way we can avoid potential typos.

mrakitin (commented):
Looks good to me. I don't have major suggestions, just a couple of minor ones.

bluesky_queueserver/server/server.py
dmgav: Let's see how the project develops. I think that at this point there are very few variables and I think only one variable is used by two application. May be it's a good idea to define additional variables that could be used instead of command-line arguments. Then variable names could be consolidated.

mrakitin (approved):

PR #151, opened 2021-04-19
Implementation of "LOOP" mode of the plan queue

dmgav:
The PR contains the following changes:

- Implementation for the LOOP mode of the queue. When the mode is enabled, successfully executed plans and instructions are pushed to the back of the queue. ZMQ and REST API and `qserver` CLI options for changing queue mode are implemented.
- Function `bind_plan_arguments`, which bind plan arguments to plan parameters based on plan description from the list of available plans.
- Renamed the environment variable `BLUESKY_HTTPSERVER_CUSTOM_MODULE` to `QSERVER_CUSTOM_MODULE`.
- REST API for uploading a batch of plans (based on the existing 0MQ API).
- Moved code `conversions.py` module from server to manager directory, since it may be used by applications other than the HTTP server.
- Moved some test code to the server directory.


danielballan (approved):

bluesky_queueserver/manager/profile_ops.py
```diff
@@ -900,6 +900,41 @@ def validate_plan(plan, *, allowed_plans, allowed_devices):
...
+    TypeError
+        Arguments could not be bound to plan parameters (raised by ``inspect.Signature.bind``.
+    """
+    param_list = copy.deepcopy(plan_parameters["parameters"])
+    parameters, _ = _construct_parameters(param_list)
+    # Create signature based on the list of parameters
+    sig = inspect.Signature(parameters)
+    # Verify the list of parameters based on signature.
+    bound_args = sig.bind(*plan_args, **plan_kwargs)
+    return bound_args
```
mrakitin, line 935: Any idea why the syntax highlighting is not working for this chunk? Can the file be read with no syntax errors?

PR #160, opened 2021-05-26
Include all items of types inherited from `OphydObject` into the list of devices

dmgav:
In the original code the list of devices was generated by picking objects that are subclasses of `ophyd.Device`. In this PR, the criteria for selection of devices is changed to include all types inherited from `OphydObject`. The list of devices will include all objects that are subclassed from `ophyd.Device` and `ophyd.Signal`. Any objects of custom types inherited from `OphydObject` will be also included.

### DIFF
--- a/bluesky_queueserver/manager/tests/test_profile_ops.py
+++ b/bluesky_queueserver/manager/tests/test_profile_ops.py
@@ -1031,7 +1031,11 @@ def test_devices_from_nspace():
     nspace = load_profile_collection(pc_path)
     devices = devices_from_nspace(nspace)
     for name, device in devices.items():
-        assert isinstance(device, ophyd.Device), f"The object '{device}' is not an Ophyd device"
+        assert isinstance(device, ophyd.ophydobj.OphydObject), f"The object '{device}' is not an Ophyd Object"
+
+    # Check that both devices and signals are recognized by the function
+    assert "custom_test_device" in devices
+    assert "custom_test_signal" in devices
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


bluesky_queueserver/manager/tests/test_profile_ops.py
```diff
@@ -1031,7 +1031,11 @@ def test_devices_from_nspace():
     nspace = load_profile_collection(pc_path)
     devices = devices_from_nspace(nspace)
     for name, device in devices.items():
-        assert isinstance(device, ophyd.Device), f"The object '{device}' is not an Ophyd device"
+        assert isinstance(device, ophyd.ophydobj.OphydObject), f"The object '{device}' is not an Ophyd Object"
```
danielballan, line 1034: As long as we are revisiting this, I think it's a good time to discard the `isinstance` check entirely and do a duck-type check instead. Could we vendor, for example, the `Readable` class proposed in https://github.com/bluesky/bluesky/pull/1446?

dmgav: It looks like Readable class will work for picking `ophyd.Signal` and `ophyd.Device` subclasses. There is also an advantage that there will be no `import ophyd` in the Queue Server code.

danielballan: Works for me.

dmgav: @danielballan I will write the issue so that we will not forget

Issue #162, opened 2021-05-31
Use `bluesky.abc.Readable` to identify devices in Run Engine Workspace


dmgav:
Currently the list of devices (detectors, motors, etc.) is generated by finding all objects in the namespace that are inherited from `ophyd.ophydobj.OphydObject`. The disadvantages of this approach:

- Works only with Ophyd objects.

- Requires importing `ophyd` (which is importing `pyepics`).

Current implementation of the function collecting devices from the workspace: https://github.com/bluesky/bluesky-queueserver/blob/5e59302c98b2435c2101868768dc1f69178ec426/bluesky_queueserver/manager/profile_ops.py#L447

The proposed change includes replacing the code 
```
       if isinstance(obj, ophyd.ophydobj.OphydObject):
```    
with
```
        if isinstance(obj, bluesky.abc.Readable):
```  
This will allow to detect all items with types inherited from `ophyd.Device` and `ophyd.Signal` without importing `ophyd`.

The class `bluesky.abc.Readable` is currently not part of Bluesky. The change may be implemented only after the PR https://github.com/bluesky/bluesky/pull/1446 is merged.

dmgav:
@danielballan @tacaswell What do we need to do before we can merge https://github.com/bluesky/bluesky/pull/1446?

danielballan:
I’ll try to review it in detail today.

ksunden:
https://github.com/bluesky/bluesky-queueserver/compare/main...ksunden:yaqc_bluesky?expand=1

For what its worth, I have a branch with this change implemented, and it has worked in my testing with non-ophyd devices (yaqc-bluesky)

I also needed to add in the Flyer interface class, since flyers are not (necessarily) Readable devices (and in fact the sim flyers in ophyd are _not_ compliant to the readable interface)

I think it may have been fixed by #160, but when I first did the test branch some tests failed due to additional items in the namespace that fit the Readable interface without being inherited from the ophyd class.

danielballan:
The protocols PR is now merged but not yet released. @ksunden Was there more you wanted to do with Flyers before we release it? I’m fine with making incremental improvements after the release…up to you.

ksunden:
I do not have any concrete thing to do about Flyers, I do not at present use them anywhere or have concrete plans to in the near term. I think @thomascobb had indicated that he had some ideas, and has done more actual thought on the matter

dmgav:
[...]

PR #163, opened 2021-06-02
ENH: use 'bluesky.abc' to identify devices in the workspace

dmgav:
Changes include the updated method identification of devices in the worker namespace. Old approach required the devices to be inherited from `ophyd.ophydobj.OphydObject`. The new approach is using `bluesky.abc` to check if the object is `abc.Readable` or `abc.Flyable`.

The PR requires https://github.com/bluesky/bluesky/pull/1446, which must be merged before this PR is merged. Unit tests are expected to pass then. 

Addresses issue https://github.com/bluesky/bluesky-queueserver/issues/162

### DIFF
--- a/bluesky_queueserver/manager/_abc.py
+++ b/bluesky_queueserver/manager/_abc.py
@@ -1,0 +1,152 @@
+# ===============================================================================================
+#   This module is temporarily used by the Queue Server code.
+#   Do not import anything from this module!
+#
+#   The code in this file is vendored from 'bluesky.abc' module. This module will be removed
+#   once the respective Bluesky PR is merged and the new version of Bluesky that supporting
+#   'bluesky.abc' is released and widely deployed. It is not intended to replace 'bluesky.abc'.
+# ===============================================================================================
+
+try:
+    from typing import Protocol, runtime_checkable
+except ImportError:
+    from typing_extensions import Protocol, runtime_checkable
+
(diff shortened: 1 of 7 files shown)


danielballan:
I think that should be sufficient, yes. I'll wait to re-review and approve once the bluesky PR is finalized and merged.

dmgav:
Considering it will take months to release and widely deploy the new version of Bluesky, I decided to temporarily create a module `bluesky_queueserver.manager._abc`, that contains copies of classes that are used for validation of devices. Queue Server will use this module if the install version of Bluesky does not contain `bluesky.abc`. Once the new version of Bluesky is deployed, the module will be removed and Queue Server will use only `bluesky.abc`.

ksunden:
At one point we suggested releasing the protocols as their own package that could be separately added as a dependency where it makes sense, as it actually does work with existing versions of bluesky itself (even if future versions of bluesky may depend on the protocols directly as well)

Perhaps we should revisit that idea if we are vendoring them here anyway, reduce duplication.

Such a package would have minimal dependencies itself (literally just the typing-extensions to make it work in versions of python prior to pep 544)

dmgav:
@ksunden I am not sure if it is worth to separate protocols into a separate package. My purpose of vendoring some code here is to make Queue Server work 'today' the same way it will work when packages with new features are released and deployed. I am planning to delete the duplicated code once the packages are deployed (I assume around September-October).

PR #172, opened 2021-07-01
Documentation on parameter validation and plan annotation

dmgav:
The PR includes first version of documentation on plan validation and parameter annotation. The placeholder files and section names are created for other topics that need to be documented.

Minor changes to the code: some of the documented API can now be imported directly from ``bluesky_queueserver`` package.


dmgav:
This PR contains mostly new text. I would suggest building the documentation locally and reading it in the browser.

bluesky_queueserver/manager/annotation_decorator.py
```diff
@@ -45,128 +45,127 @@
...
+                        #   allows to pass one plan from the group 'Plan1' or a list of
+                        #   plans from 'Plan2'.
+                        "Plan1": ("count", "scan", "gridscan"),
+                        "Plan2": ("more", "plan", "names"),
+                    },
+                },
+
+                "devices": {
+                    "description": "Parameter that accepts the list of devices.",
+                    "annotation": "typing.List[Device]",
```
callumforrester, line 137: An idea that may or may not be in-scope for this PR: Rather than use the concrete Ophyd base classes, consider referring to devices using the protocols that @ksunden defined in https://github.com/bluesky/bluesky/pull/1446. They are runtime-checkable so validation at runtime is possible e.g.
```python
if isinstance(some_device, Readable):
    ...
```

dmgav: @callumforrester Thank you for pointing out the issue. Name `Device` here is not related to Ophyd Device class name, therefore it is misleading to use it in the example (in practice you can use `Device` name). I replaced it with `DeviceType1`, so that it is clear that the name can be arbitrary. We do not pass device objects to plans over 0MQ. Instead we pass object names and then replace the names with references to objects before running the plan. So validation checks that the device names are correct. Validation will also work with lists, dictionaries, lists of dictionaries, etc. of device names. In this example the parameter value must be a list of device names, where device names may be `"det1"`, `"det2"` or `"det3"`. The worker namespace is expected to contain the devices `det1`, `det2` and `det3`.

Queue Server is already using the protocols to find devices in the namespace https://github.com/bluesky/bluesky-queueserver/blob/0fb73b701cd330f2709f86c19bc46a0e45c16b4d/bluesky_queueserver/manager/profile_ops.py#L449-L452 and determine device properties https://github.com/bluesky/bluesky-queueserver/blob/0fb73b701cd330f2709f86c19bc46a0e45c16b4d/bluesky_queueserver/manager/profile_ops.py#L1602-L1609 so it will work with objects other than Ophyd devices. Queue Server code has temporary copies of `Readable`, `Writable` and `Flyable` classes (https://github.com/bluesky/bluesky-queueserver/blob/main/bluesky_queueserver/manager/_protocols.py). The temporary copies will be removed when the new version of Bluesky is deployed.

PR #183, opened 2021-07-22
ENH: show helpful error message if no redis server available

developingAlex:
Hi Dmitri, I set up the queueserver on a new machine and forgot to provide dependencies like redis so I thought this might be a small help if another person gets stuck on that, what do you think?

### DIFF
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -223,12 +223,15 @@ class PlanQueueOperations:
             async with self._lock:
                 try:
                     self._r_pool = await aioredis.create_redis_pool(f"redis://{self._redis_host}", encoding="utf8")
-                except OSError:
-                    logger.error(
-                        "Exception in attempting to create the redis pool, "
-                        f"is a redis server available at {self._redis_host} ?"
+                except OSError as ex:
+                    error_msg = (
+                        f"Failed to create the Redis pool: "
+                        f"Redis server may not be available at '{self._redis_host}'. "
+                        f"Exception: {ex}"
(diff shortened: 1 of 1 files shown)


dmgav:
@developingAlex Thank you for pointing out the issue. It is helpful to have clearer message in case Redis is not installed. I rearranged the code a little bit. If you agree, we can merge this PR.

developingAlex:
Hi Dmitri thanks for your review and suggested changes,

I agree, my only reservations was that I feel its very verbose with the exception stack trace logging when we know that the cause is just that it couldn't connect to redis.

And this is compounded by the fact that it gets repeatedly printed out just about every second until a redis server does become available because the watchdog process keeps reviving it after it crashes.

But I don't know if I would really change anything because the redis availability is really crucial to the whole thing so it is correct that it crashes if it can't connect.

Likewise it's also correct for the watchdog to keep reviving it because it could just be that it was started up before redis became available and when redis does become available we'd want the queueserver to still be there and trying to connect.

Also I'm not sure if we can be truly certain that any time that redis code throws an exception that we can assume it's due to failing to connect, I mean what if redis was actually up and available and it was something else unexpected that was raising the exception? In that case I'd want to see the stack trace.

So I think this is OK for the time being! Happy for you to merge it if you're still happy with it!

Thanks!

dmgav:
Hi Alex,

I am thinking about changing the behavior of the queue server so that it does not attempt to restart the manager every 5 seconds if it the manager fails to start the first time. The manager typically fails to start due to some issue that does not go away and repeatedly printing the same error messages is annoying. Re-raising exceptions seems to preserve stack trace.

PR #187, opened 2021-08-06
Avoid conflation of documents-publishing and console-publishing.

danielballan:
Renamed CLI parameters of `start-re-manager`:

- `--zmq-publish` is renamed to `--zmq-publish-console`
- `--zmq-publish-addr` is renamed to `--zmq-publish-console-addr`


dmgav:
I think changes to parameter are reasonable, except that the parameter names may become too long. Setting `det.kind` in the simulated startup files is fine, but there should be some comment on why it is needed here and where it is used, otherwise nothing prevents someone from deleting it in the future (CI will still pass).

Also, since command parameters are changed, tests and documentation needs to be updated accordingly. I probably can update it later.

danielballan:
Oops, I didn't mean to commit the `det.kind` change. I think we should _not_ commit that actually go ahead with a separate profile.

danielballan:
I am also happy to clean this up myself after I return from vacation if you don't get to it first.

dmgav:
I think I made all necessary changes. The PR is ready to be merged.

PR #189, opened 2021-08-30
Stop queue on unresolved pause

developingAlex:
Hi Dmitri,

For your consideration, I have a pull request that changes the behaviour of the queueserver around (deferred) pausing the RE Worker, 
Please see the commit comment for the details,

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -385,6 +385,7 @@ class RunEngineManager(Process):
                     self._worker_state_info = ws
                     if ws["re_state"] == "paused":
                         self._re_pause_pending = False
+
                     if self._manager_state == MState.CLOSING_ENVIRONMENT:
                         if ws["environment_state"] == "closing":
                             self._fut_manager_task_completed.set_result(None)
@@ -397,7 +398,7 @@ class RunEngineManager(Process):
                         if ws["environment_state"] == "closing":
                             self._fut_manager_task_completed.set_result(False)
 
-                    elif self._manager_state == MState.EXECUTING_QUEUE:
+                    elif self._manager_state in (MState.EXECUTING_QUEUE, MState.PAUSED):
                         if ws["re_report_available"]:
                             self._loop.create_task(self._process_plan_report())
 
@@ -2068,9 +2069,12 @@ class RunEngineManager(Process):
             if self._worker_state_info:
                 item_uid_running = self._worker_state_info["running_item_uid"]
                 re_state = self._worker_state_info["re_state"]
+                re_report_available = self._worker_state_info["re_report_available"]
                 re_deferred_pause_requested = self._worker_state_info["re_deferred_pause_requested"]
-                if item_uid_running:
-                    # Rare case when the manager was restarted while the deferred pause was pending.
+                if item_uid_running and (re_report_available or (re_state != "idle")):
+                    # If 're_state' is 'idle', then consider the queue as running only if
+                    #   there is unprocessed report. If report was processed, then assume that
+                    #   the queue is not running.
                     self._re_pause_pending = re_deferred_pause_requested
                     # Plan is running. Check if it is the same plan as in redis.
                     plan_stored = await self._plan_queue.get_running_item_info()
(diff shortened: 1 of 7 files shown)


bluesky_queueserver/manager/manager.py
```diff
@@ -495,14 +498,23 @@ async def _start_plan_task(self, stop_queue=False):
...
             logger.info(err_msg)
 
         elif self._queue_stop_pending or stop_queue:
             self._manager_state = MState.IDLE
+            self._worker_pause_pending = False
             success, err_msg = False, "Queue is stopped."
             logger.info(err_msg)
 
+        elif self._worker_pause_pending:
+            self._worker_pause_pending = False
```
dmgav, line 512: I guess this line is redundant. It is repeated two lines below.

dmgav (commented):
Thanks @developingAlex I completely missed this detail. It is an interesting question. I guess that if user is attempting to pause the plan, it may be reasonable to assume that in most workflows the queue should be stopped if currently running plan can not be paused, but there could be applications when pause request should be ignored. May be `re_pause` API should have additional optional parameter that tells whether the queue should be stopped (default behavior) or pause request should be ignored? 
@tacaswell What is your opinion?

We will also need to write tests to make sure that functionality continues to work in the future.

bluesky_queueserver/manager/manager.py
developingAlex: woops! yes you're right, I will remove it

dmgav:
I would like to summarize the issue.

Assume that Queue Server receives deferred `re_pause` request after the last checkpoint of the plan. The plan can not be paused, so it will run to completion. How should the Queue Server handle the request?

List of choices:

1. Ignore the request (current behavior). The request is passed to Run Engine, but the plan still runs to completion and the next plan is started.

2. Stop the queue after completion of the current plan (proposed in this PR).

3. Option 1 is default, but option 2 can be enabled by passing additional parameter to `re_pause` request.

4. Option 2 is default, but option 1 can be enabled by passing additional parameter to `re_pause` request.

tacaswell:
I think option 2 is the best option.  I am worried about user surprise at "I asked it to pause AND IT KEPT GOING" in the case where they got un-lucky and missed the last checkpoint.  I see how the current behavior is technically correct (you said pause the running plan, I tried, you missed your window 🤷🏻 ).  It is also _mostly_ consistent with what would happen if you did

```
def multi_plan(...):
    yield from planA()
    # <- pause happens here after last checkpoint of A but before first checkpoint of B
    yield from planB()
```

One could make a weak-case that the pause should apply to the next plan in the queue, but I think that if we are going to special case handling a un-handled pause we might as well stop the queue rather than letting the next plan _start_ and then pause.

For 3 and 4 I am not convinced that the current behavior is something that people will actually want to opt-into as, from a user point of view "I said pause!  I the human want to think a bit at human speeds before you the computer go on taking maybe wrong/useless data at computer speeds." is the intention, not "I said pause unless I lost a timing race with the computer and then it should carry on".  

----

Talking to @dmgav I proposed collapsing all of the `re_xyz` messages into two where `re_pause` would take as optional input `({deferred,immediate}, {pause, stop, abort, halt}, {queue stop, queue_continue})` and only `re_resume` to take it out of an actually paused state.  This would reduce functionality a bit, but I think that it would remove enough complexity to be worth it.  Independent of if this is a reasonable position or not, it should not be handled in this PR.

dmgav:
@tacaswell Thanks for comments. I remember, the initial implementation of API had 'collapsed' implementation for `stop`, `resume`, `abort` and `halt` and it was decided to break it into a set of separate API. As for passing `stop`, `abort` and `halt` with the call, I believe that in many cases users pause the plan to inspect equipment and they are not sure what to do next. Similar to IPython workflow, operation of pausing a plan and then stopping/aborting/halting or resuming it are two separate operations.

As for this PR, I agree with option #2.

developingAlex:
Thanks for the feedback @dmgav @tacaswell 

Please let me know if there's any work I would still need to add to this?

I wasn't sure how to write a test for it but I did test it just locally

dmgav:
@developingAlex I pushed some commits to this PR. If you agree with the changes, we can merge this PR.

PR #198, opened 2021-10-18
Allow for requesting of device configuration

ksunden:
This serves as a concrete implementation of a proposal I brought up on the Nikea slack regarding allowing access to device configuration.

This serves as a _read-only_ access to configuration of devices in the worker namespace.
Such devices can provide things like units or limits which can help make UIs to interact with the queueserver with proper gaurdrails that present a streamlined user experience.
For many devices (at least from what I can tell) these units and limits are already defined.

This was a quick and dirty proof of concept implementation, happy to revise if needed.

The intent here is to allow for transfer of (potentially dynamic) information via the queueserver zmq interface without opening up the can of worms that is a general and extensible RPC layer.

Other factors to consider:
- I do no checking of whether the requesting user is _allowed_ to see the device, perhaps that is a desirable check
- Current (and as far as I can tell, any) implementation does require the worker environment to be open, as the actual `read_configuration` call happens there as that is the owner of the devices

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -977,6 +977,29 @@ class RunEngineManager(Process):
             "devices_allowed_uid": self._allowed_devices_uid,
         }
 
+    async def _device_configuration_handler(self, request):
+        """
+        Returns the list of allowed devices.
+        """
+        logger.info("Returning the device configuration ...")
+
+        try:
+            supported_param_names = ["device"]
+            self._check_request_for_unsupported_params(request=request, param_names=supported_param_names)
+
+            device_configuration = await self._comm_to_worker.send_msg("request_device_configuration", {"device": request["device"]})
+            success, msg = True, ""
+        except Exception as ex:
+            device_configuration = {}
+            success, msg = False, str(ex)
+
+        return {
+            "success": success,
+            "msg": msg,
+            "device_configuration": device_configuration,
+        }
+
+
     async def _permissions_reload_handler(self, request):
         """
         Reloads the list of allowed plans and devices and user group permission from the default location
@@ -1972,6 +1995,7 @@ class RunEngineManager(Process):
             "queue_get": "_queue_get_handler",
             "plans_allowed": "_plans_allowed_handler",
             "devices_allowed": "_devices_allowed_handler",
+            "device_configuration": "_device_configuration_handler",
             "permissions_reload": "_permissions_reload_handler",
             "history_get": "_history_get_handler",
             "history_clear": "_history_clear_handler",
--- a/bluesky_queueserver/manager/qserver_cli.py
+++ b/bluesky_queueserver/manager/qserver_cli.py
@@ -49,6 +49,8 @@ qserver allowed plans          # Request the list of allowed plans
 qserver allowed devices        # Request the list of allowed devices
 qserver permissions reload     # Reload the list of allowed plans and devices and user permissions from disk
 
+qserver device configuration <device>  # request the configuration for a device, environment must be open
+
 qserver queue add plan '<plan-params>'                 # Add plan to the back of the queue
 qserver queue add instruction <instruction>            # Add instruction to the back of the queue
 qserver queue add plan front '<plan-params>'           # Add plan to the front of the queue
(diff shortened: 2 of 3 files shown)


dmgav:
Unfortunately, I am more and more inclined to support original idea that there should be a separate channel for communication with devices (monitoring, direct control etc.) not related to the Queue Server. The fact that no one worked on it does not mean that it can not or will not be done, since it is an expected part of the system architecture and there are requests for remote monitoring features. There is a lot of opposition to idea of general purpose RPC, since it provides a lot of ways to break the system and warnings to use advanced features sparingly do not work. I am hesitant about specialized API like this one, because they serve only a single purpose and they require the same infrastructure as general purpose RPC. They also provide lots of ways to break the system in worse ways (e.g. an Ophyd object that has `read_configuration()` method that do some computations, move a motor or do things it is not supposed to do). 

@ksunden I don't completely understand your control system, so it is hard to tell if separate channel will work in your case. I believe you do not use Epics to communicate with physical devices. Do you still communicate with devices over the network?

ksunden:
> (e.g. an Ophyd object that has read_configuration() method that do some computations, move a motor or do things it is not supposed to do)

Is this really a usecase we are worried about? Is there an example of when this is a reasonable thing to do? Certainly it is _possible_, but I'd argue that it flies in the face of the semantics of a "read" operation.

> Do you still communicate with devices over the network?

The short answer is yes, we _could_ acquire this information through the devices directly...

However, this gets into configurational complexity: the set of devices I may wish to inquire about is precisely those that the queueserver already knows about and is interfaced with, without my client having any other reason to contact the devices directly. The client may be running remotely (I include 'in the office the next room over' as "remote"), and would prefer to be able to avoid/limit direct remote connections to hardware if possible. I would like to avoid a scenario which requires multiple machines to be updated each time there is a configuration change as well as make it easier to switch remote machines to several different experimental systems.

The primary pieces of information I would like to present to the user who is enqueuing a plan are:

- limits of a device specified by another parameter of the plan (and therefore not known by the plan itself)
   - Limits for a given device _can and do_ change over the course of an experimental session, so fully static "ask once and cache" is not necessarily the right call, though usually _not_ without halting the queue
- units of a device specified by another parameter of the plan (I have unit aware plans, wish to limit drop down options to relevant units)

Both of these are commonly found (when defined, at least) in `describe` for the device.

Secondary info, which would be nice and could go in `describe` or `read_configuration`:
- options for a parameter of some (in particular tuning) specialized plans that differ based on selected device
   - These are relatively reasonable to get heuristically from the list of devices as a whole for what I have in mind, so if this was more difficult, that could be set aside.


Scans in my lab are regularly multiple hours long and interrupting them to acquire information would be very frustrating. Admittedly, some features (such as live plotting) already do not work if you started the application after the run started, as it uses information from the start/descriptor docs to format the plot and it cannot receive those after the run is already going.

untzag:
@ksunden's comment about start documents makes me realize that we _could_ reach into databroker and find the most recent describe document to get this information. Currently our application doesn't interact with databroker at all, but it would be fairly easy to add that functionality.

Alternatively, perhaps we could cache the most recent start and by-stream describe documents and offer those cached values through the queueserver API. Just to keep the networking complexity as simple as possible. Perhaps there would be other documents worth caching.

It's fairly easy to handle the case of "new queueserver, nothing known", we can enqueue a null plan just like the client starts the run engine if none is running.

untzag:
Perhaps zmq proxy or kafka (?) could automatically cache those documents. I don't know the data model of either.

dmgav:
My point was that implementing general RPC is not significantly decreasing security of the system. General RPC may be slightly more complicated, but it could be better tested and work more reliably as a result. It can also be used to implement arbitrary application-specific functionality if needed (at user's risk).

One of the ideas is that all queue operations should be functional before the worker environment is opened. The environment should be created only when it is time to execute the queue. There are situations when the environment can not be created because startup files technically can not be loaded, or when operation of creating the environment should be disabled (this is not implemented yet, but it must be implemented before production deployment). Availability of the worker environment should not influence ability to edit the queue in any way. Also if queue operations do not require access to the environment, it will be straightforward to 'virtualize' the queue in the multiuser system, where user's queues reside on a web server and loaded to the queue server only when the user is given access. The separate monitoring channel is expected to have security features that allow it to run 24/7 without issues. 

We are planning to use kafka for all queue server deployments. Kafka will retain all documents for the run. There is some code that does live plotting based on streamed documents from kafka in `bluesky-widgets`. There were plans to implement framework for live plotting, which allows to display data starting from the beginning of the run even if application is started in the middle of the run, but I am not sure if it was completed. All documents are retained by kafka and an application can receive the stream starting from the beginning of the run at any time.

untzag:
> All documents are retained by kafka and an application can receive the stream starting from the beginning of the run at any time.

Great, it seems like @ksunden and I can get what we want without any changes then. We just have to use kafka instead of zmq-proxy. If you have the time could you provide a minimal example of a Python client that does this?

dmgav:
Here is example of Kafka consumer that saves documents to Mongo: https://github.com/bluesky/bluesky-kafka/blob/master/bluesky_kafka/mongo_normalized_consumer.py
There are also some examples in Bluesky Widgets repository, for example the consumer that saves thumbnails of data plots:
https://github.com/bluesky/bluesky-widgets/blob/2f3dfa1901734bc0d85a1519ec69adc5d5d77562/bluesky_widgets/examples/kafka_figures.py

ksunden:
Closing as we decided on another route (namely ksunden/bluesky-hwproxy)

PR #221, opened 2022-03-07
PERF: Disable plots in best effort callback

ksunden:
## Description
Disable plots as they are not used anyway

## Motivation and Context

Plotting incurs a significant (~150 ms per read, more if you have more channels) time delay that serves no real purpose as the queueserver is running headless and does not save the plots to images either.

This still has the live table (with BEC determined columns).

### Changed
- No longer plot with best effort callback (improves performance)


dmgav (approved):

dmgav:
Thanks, @ksunden

PR #257, opened 2022-09-22
remove vendored protocols

untzag:
## Description

Remove vendored protocols, instead relying on protocols module within bluesky.

Pinned to bluesky>=1.7.0, which is now more than one year old.

## Motivation and Context

This change was always intended, now it can be completed since protocols are now mature and widely distributed.

The protocols within queueserver were diverging from the correct protocols within bluesky.

### Changed
- now requires bluesky>=1.7.0

### Removed
- removed built-in protocol support, now relies on bluesky.protocols

## How Has This Been Tested?

Any problems should be caught by existing tests.


danielballan (approved):

mrakitin (approved):
Good idea to stop vendoring the protocols.

PR #258, opened 2022-09-22
test against Python 3.10

untzag:
We already use Python 3.10, but we don't yet test against it.


dmgav:
I see that all packages for Python 3.10 are available now. I will take a look on why some tests fail. There could be some minor incompatibilities.

dmgav:
I pushed some minor change to fix compatiblity issue.

untzag:
Thanks for jumping in and fixing this @dmgav

PR #266, opened 2023-03-31
ENH: Auto-start queue as non-default option 

maffettone:
## Description

This extends the enum for `MState`,  adds an attribute to the `REManager`, and some methods for interfacing with that attribute via ZMQ. It provides the user an option for an empty queue, that is otherwise uninterrupted, to remain "hot" and auto-start when new items are added. This retains the default behavior of the `REManager`, but creates a new type of 'idle' state beyond `MState.IDLE`. All other interruptions (i.e. stopping the queue, an error in a plan, etc.) will stop the queue in an `IDLE` state, that will require user intervention to restart the queue. This will for example, allow a user to ask for a stop to the queue during the final running plan, and not have any new added plans automatically run. 

## Motivation and Context

Beamline scientists have been begging for this. It reflects the behavior that they are accustomed to with the RunEngine (prompt execution of plans), and enables remote agents to drive an experiment starting from an empty queue. While there is already functionality for prompt execution of plans, this would enable the addition of a plan or set of plans from multiple sources, with automatic processing from the queue. I broadly agree that this should not be the default behavior, but should be an option. 

## Summary of Changes for Release Notes

Adds an optional autostart state to `REManager`, so that an empty queue can begin processing plans automatically as they are added. 

### Fixed
- Adjusts the `_start_plan_task` to first check for a queue stop, then check if there are pending plans. This ordering wasn't an issue when both no pending plans and a queue stop triggered similar behavior. 

### Added
- Additional state to `MState`
- Activate and deactivate functions for autostart, with handlers

### Changed
- Caveat to some checks for `IDLE` to optionally change to `AUTOSTART` instead. 
- 
## How Has This Been Tested?

Written unit tests following the same syntactical approach in `test_scenarios.py` and `test_zmq_api.py`.

### DIFF
--- a/bluesky_queueserver/manager/manager.py
+++ b/bluesky_queueserver/manager/manager.py
@@ -831,16 +831,16 @@ class RunEngineManager(Process):
             else:
                 logger.info("No items are left in the queue.")
 
-            if not n_pending_plans:
-                self._manager_state = MState.AUTO_START if self._queue_auto_start else MState.IDLE
-                self._re_pause_pending = False
-                success, err_msg = False, "Queue is empty."
-                logger.info(err_msg)
-
-            elif self._queue_stop_pending or stop_queue:
+            if self._queue_stop_pending or stop_queue:
                 self._manager_state = MState.IDLE
                 self._re_pause_pending = False
                 success, err_msg = False, "Queue is stopped."
+                logger.info(err_msg)
+
+            elif not n_pending_plans:
+                self._manager_state = MState.AUTO_START if self._queue_auto_start else MState.IDLE
+                self._re_pause_pending = False
+                success, err_msg = False, "Queue is empty."
                 logger.info(err_msg)
 
             elif self._re_pause_pending:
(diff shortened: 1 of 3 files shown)


maffettone:
This is failing on 
`bluesky_queueserver/manager/tests/test_scenarios.py::test_large_datasets_03`
`bluesky_queueserver/manager/tests/test_zmq_api.py::test_zmq_api_queue_item_add_batch_1`
`bluesky_queueserver/manager/tests/test_zmq_api.py::test_zmq_api_queue_item_add_batch_2``bluesky_queueserver/manager/tests/test_zmq_api.py::test_zmq_api_queue_item_add_batch_1`

dmgav:
At this point I am not sure it's a good idea to make 'autostart' a separate manager state, because it does not reflect what is the manager is currently doing. It looks more like a queue mode. I will think where to put it when I get back from vacation.

dmgav:
Closing in favor of PR https://github.com/bluesky/bluesky-queueserver/pull/269

PR #276, opened 2023-06-08
Shift from aioredis to redis

clintonroy:
Aioredis is now, more or less, merged into redis. As aioredis is now archived, better to shift to redis directly.

## Description
Modify imports to use redis. Remove code that handles older versions of aioredis.

## Motivation and Context

Attempting to shift to python3.11 showed up a couple of issues with aioredis, and that led me down the path of discovering that it's now archived.

## How Has This Been Tested?

We've tested this, in a limited way, on AS beamlines. We can start the queueserver, add and run plans.

I could not run the tests by following the contributing guide, hoping the CI does that.

### DIFF
--- a/bluesky_queueserver/manager/plan_queue_ops.py
+++ b/bluesky_queueserver/manager/plan_queue_ops.py
@@ -5,13 +5,10 @@ import logging
 import time as ttime
 import uuid
 
-import aioredis
+import redis.asyncio
 from packaging import version
 
 logger = logging.getLogger(__name__)
-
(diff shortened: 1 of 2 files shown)


tacaswell:
This seems reasonable to me.  The failure looks transient (given that it only happens in 1 test case....), tried restarting.

dmgav:
Looks good to me. Unfortunately the tests are complicated and there are a lot of them, so failures on CI are quite common. 

@clintonroy What issues did you have with running the tests locally?

dmgav (approved):

PR #281, opened 2023-07-19
Update installation instructions: supported Python versions

dmgav:
Currently Python 3.7-3.11 are supported. CI tests are run for Python 3.9-3.11. Support for Python 3.7 will be dropped soon.


dmgav:
CI fails due to incompatibility with Pydantic v2, which will be the subject of the next PR.

PR #297, opened 2024-02-26
Include device `.name` in API data

flowln:
This field can be used to show a better user-facing name in a GUI.

To be honest, maybe the field should be named something else, like `display_name` or `device_name`, but since neither Bluesky nor Ophyd constrains what `name` can be, I used the generic `name`. Tell me if you think it should be something else!

## Motivation and Context
This allows us to have a potentially ugly Python variable name, while having a nicer name attribute on the device, that users can interact with in a GUI application.

### Added
A `name` field to the device list in the API, corresponding to the device's `.name` attribute, when it exists.

### DIFF
--- a/src/bluesky_queueserver/manager/profile_ops.py
+++ b/src/bluesky_queueserver/manager/profile_ops.py
@@ -3297,7 +3297,7 @@ def _prepare_devices(devices, *, max_depth=0, ignore_all_subdevices_if_one_fails
     from bluesky import protocols
     from ophyd.areadetector import ADBase
 
-    def get_device_params(device):
+    def get_device_params(device, device_obj_name):
         movable_protocols = (protocols.Movable,)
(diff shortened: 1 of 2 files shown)


dmgav:
Thank you for contributing to the project. We consider adding user-friendly name to the device description as a useful feature. But the primary purpose of the `name` attribute of the Ophyd object is different, and using it for arbitrary user-friendly name is not the best practice. There is an issue in Ophyd project, proposing a new attribute `display_name` for `OphydObject` class: https://github.com/bluesky/ophyd/issues/1161 Once implemented in Ophyd, the `display_name` could be included in the device description.

dmgav:
I am converting the PR to draft for now.

dmgav (approved):

dmgav:
Since the work on implementation of a long name seems to be delayed, we can use the value of the the device `.name`, which can be different from the respective variable name.

Issue #305, opened 2024-08-05
Bluesky plan_stubs not recognized with Bluesky v1.13.0a4


canismarko:
## Expected Behavior

If I import bluesky plan stubs in my queueserver startup file, they should be available in the queueserver as plans. This was the behavior when using bluesky v1.13.0a3 and below.

## Current Behavior

The plan stubs are not available.

## Possible Solution

Queueserver uses ``inspect.isgeneratorfunction()`` to determine if a token corresponds to a valid plan. With bluesky v1.13.0v4, plan stubs are now wrapped and no longer pass this check.

@tacaswell's suggestion is to include a separate check to detect these wrapped plan stubs: https://github.com/bluesky/bluesky/issues/1779#issuecomment-2267222073

## Steps to Reproduce (for bugs)

1. Install Bluesky v1.13.0a4
2. Write a queuerserver startup file that imports ``from bluesky.plan_stubs import mv``
3. Start the queueserver
4. Look at the list of available plans.

## Context

We use the queueserver for executing scans at our beamline. We include a few plan stubs for making adjustments to the beamline in-line in the queue.

## Your Environment

https://gist.github.com/canismarko/39d39055a71b165faba801872162f7ba

canismarko:
Fixed: https://github.com/bluesky/bluesky/pull/1780

dmgav:
Resolved in https://github.com/bluesky/bluesky-queueserver/pull/306

canismarko:
I've been using this successfully for the past few weeks, so I think everything is good. Closing this issue now. Thank you!

PR #306, opened 2024-09-07
Use is_plan from bluesky utils

jwlodek:
## Description
Switch to use new `is_plan` function from bluesky utils.

## Motivation and Context
Solves #305 . Requires https://github.com/bluesky/bluesky/pull/1780 to be released, and pinned in the `requirements` file.

## Summary of Changes for Release Notes
Remove `is_plan` function, use instead the equivalent from bluesky utils that can handle the decorated functions.

### Fixed
Bluesky builtin plans that use the `@plan` decorator are now correctly identified by queueserver.

### DIFF
--- a/bluesky_queueserver/manager/profile_ops.py
+++ b/bluesky_queueserver/manager/profile_ops.py
@@ -24,6 +24,8 @@ import yaml
 from numpydoc.docscrape import NumpyDocString
 from packaging import version
 
+from bluesky.utils import is_plan
+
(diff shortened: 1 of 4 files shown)


dmgav:
In my opinion, the internally defined function `is_plan` should not be deleted, but instead modified to call the function defined in bluesky. I would also add some tests to make sure the plans are actually recognized.

dmgav (approved):

PR #308, opened 2024-10-14
MNT: pass callbacks directly to _RE.__call__

tacaswell:
## Description

The second positional argument to RE() is subscriptions to processed for that invocation.  This avoids adding one more layer to the callstack and a possible confusion with plans that do not set their own name being reported as having the `'plan_name'` of `'subs_wrapper'` in the start document.

## Motivation and Context

This caused some confusion at HEX.  The ultimate fix is that plans that generate start documents should insert their own name into the meta-data of the `open_run` plan so that when used from inside of a bigger plan the name is stable.  This change will not fix the ultimate problem, but it will change what the user sees to be the name of the plan they wrote so it will be clearer what is going on and what they have to do to fix it.

## How Has This Been Tested?

Expect all tests to pass after this change and there to be no meaningful user-facing change.

### DIFF
--- a/bluesky_queueserver/manager/worker.py
+++ b/bluesky_queueserver/manager/worker.py
@@ -463,8 +463,6 @@ class RunEngineWorker(Process):
         plan_info = parameters
 
         try:
-            from bluesky.preprocessors import subs_wrapper
-
             with self._allowed_items_lock:
                 allowed_plans, allowed_devices = self._allowed_plans, self._allowed_devices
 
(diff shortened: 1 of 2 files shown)


mrakitin:
Is it this warning which causes the documentation to fail?
> WARNING: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.

mrakitin (approved):

mrakitin:
We should test it in the lab before merging.
cc @jwlodek

PR #309, opened 2024-10-14
shift requirements to dev

prjemian:
Both pyarrow and pandas were added as package requirements in commit  672f579 as part of PR#301.  Neither of these are imported by this package.  Shift the both of them to `requirements-dev.txt`.

## Motivation and Context
We ran into an installation dependency in https://github.com/prjemian/model_instrument/pull/5#issuecomment-2408085385.  On inspection, this seems not to be a real requirement of this package.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,3 +22,5 @@ sphinx
 sphinx_rtd_theme
 # Extra dependencies for development
 fastapi[all]
+pandas
+pyarrow
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,5 +22,8 @@ sphinx
 sphinx_rtd_theme
 # Extra dependencies for development
 fastapi[all]
+# moved from requirements.txt
 pandas
 pyarrow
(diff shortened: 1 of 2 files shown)


tacaswell:
#308 has a fix for the doc failure.

tacaswell:
I suspect a few more can go (like mpl, skimage, and databroker).

prjemian:
Let's get these through.  Probably enough for the PR that triggered this.  Any others can be noted in a new issue/branch/PR.

dmgav:
For the future, the following code was added at the very start of the project, serves no practical purpose and should be removed. The respective parameters should be removed. The instance of `RE` is always created in the startup code as well as all subscriptions. Built-in features only confuse beginners who look at the examples.
https://github.com/bluesky/bluesky-queueserver/blob/d529fcb312c6840877627ab3e44a062f55e184a2/bluesky_queueserver/manager/worker.py#L1478-L1536

dmgav (approved):

prjemian:
@dmgav , @tacaswell -- Thanks!

PR #313, opened 2024-11-25
Dev/make kafka optional

tacaswell:
## Description

Buries the bluesky-kafka import to where it is actually used.

## Motivation and Context

Lighten the default installation weight.

### Changed

- made bluesky-kafka an optional dependency
- added "kafka", "dev", and "all" extras_requires groups


dmgav:
Kafka dependency was removed in https://github.com/bluesky/bluesky-queueserver/pull/322

Issue #314, opened 2025-01-27
IPython Kernel Configuration in ipython_kernel_config.py breaks `worker`


kivel:
## Description
The IPython profile configuration for the kernel found in `<ipython_dir>/<profile_dir>/ipython_kernel_config.py` is being ignored. Specifically, settings such as:

```python
c.ConnectionFileMixin.connection_file = "/tmp/kernel-tst-fixed-ports.json"
c.ConnectionFileMixin.ip = "10.17.0.113"
c.ConnectionFileMixin.control_port = 60001
c.ConnectionFileMixin.hb_port = 60002
c.ConnectionFileMixin.iopub_port = 60003
c.ConnectionFileMixin.shell_port = 60004
c.ConnectionFileMixin.stdin_port = 60005
```
are being replaced by random ports instead of the specified values.
This causes the environments fail to load consistently.

## Why is this problematic?

**Firewall Restrictions**: If only certain ports are open in a firewall, using `ipython_kernel_config.py` allows binding the kernel to fixed ports, ensuring connectivity.
**Deterministic Connection**: By defining a deterministic connection file, remote connections (e.g., jupyter console and jupyter lab) can reconnect after the environment is closed and re-opened.

## Expected behavior
The IPython kernel should respect the configurations specified in `ipython_kernel_config.py`. This includes binding to the specified IP address and ports, and using the defined connection file. This behavior ensures that users can manage their network configurations effectively and maintain stable remote connections.

#### connection file only

```python
# ---  ~/.ipython/profile_dynamic_config/ipython_kernel_config.py
# Configuration file for ipython-kernel.
from pathlib import Path

KERNEL_PATH = "/tmp/jupyter/kernels"
BEAMLINE = "TST"
# Create the path
connection_file = Path(KERNEL_PATH) / f"kernel-{BEAMLINE.lower()}-dynamic.json"

# Ensure the folder exists
connection_file.parent.mkdir(parents=True, exist_ok=True)

c = get_config()  #noqa

c.ConnectionFileMixin.connection_file = str(connection_file)
```

```shell
start-re-manager --use-ipython-kernel=ON --startup-profile=dynamic_config
```

In case this is the first time the profile is used, or the connection file has been manually removed, the environment will open and the connection file will be created.
However, when the environment is closed and re-opened, the worker will attempt to create a new ipython kernel, with newly assigned random ports. This will raise a ValueError exception:
```shell
[...]

PR #315, opened 2025-01-27
Fix IPKernel config handling

kivel:
# Fix errors arising from disregarding the `ipython-kernel-config.py`

## Description
Providing a kernel configuration makes the re-manager fail to load the environment with an ipython kernel.
See #314 for details.

## Motivation and Context
When using an ipython kernel, users will expect the provided configuration for the kernel in the profile directory to be obeyed.
The bug resulted from explicitly defining a kernel-client with connection information produced in scope, rather than providing the IPKernelApp with the respective overwrites and creating the client from the KernelApp.

### Fixed
No direct fixes as the expected behavior can be achieved by arranging the code.

### Added
- the `find_kernel_ip` function is made a little more robust in case the socket calls are failing

### Changed
- the sequence in which the `IPKernelApp` and the `BlockingKernelClient` are created are reversed.
- the `BlockingKernelClient` is obtained via the provided method from IPKernelApp `self._ip_kernel_client = self._ip_kernel_app.blocking_client()`

### Removed
- `self._ip_connect_file` removed because is was never used.
- `self._ip_connect_info` was removed because there is no real reason to keep this state, the info can be obtained from by calling `self._ip_kernel_app.get_connection_info()`
- the calles to `generate_random_port(kernel_ip)` have been removed, as the IPKernelApp will used random ports be default, or read the ports from the config or an existing connection file.

## How Has This Been Tested?
The configurations for two different profile setups are provided in issue #314.
Both cases work as expected.
Additionally, invoking `start-re-manager --use-ipython-kernel=ON` with a profile that doesn't bear a configuration, will behave identically to the existing implementation.

### DIFF
--- a/bluesky_queueserver/manager/worker.py
+++ b/bluesky_queueserver/manager/worker.py
@@ -1797,10 +1797,10 @@ class RunEngineWorker(Process):
                 """Find the IP address to use for the IPython kernel."""
                 if ip_str == "localhost":
                     return "127.0.0.1"
-                    
+
                 if ip_str != "auto":
                     return ip_str
-                    
+
                 # Try to determine IP automatically
                 try:
                     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
@@ -1808,7 +1808,7 @@ class RunEngineWorker(Process):
                     ip_address = sock.getsockname()[0]
                     sock.close()
                     return str(ip_address)
-                    
+
                 except Exception as e:
                     logger.error("Error determining IP automatically: %s", str(e))
                     return "127.0.0.1"  # Fallback to localhost
@@ -1837,7 +1837,6 @@ class RunEngineWorker(Process):
 
                 self._ip_kernel_monitor_always_allow_types = ["error"]
                 self._ip_kernel_monitor_collected_tracebacks = []
-
 
                 logger.info("Initializing IPython kernel ...")
                 self._ip_kernel_app.initialize([])
--- a/bluesky_queueserver/manager/worker.py
+++ b/bluesky_queueserver/manager/worker.py
@@ -1826,10 +1826,11 @@ class RunEngineWorker(Process):
             if self._success_startup:
 
                 def start_jupyter_client():
-                    from jupyter_client import BlockingKernelClient
+                    # from jupyter_client import BlockingKernelClient
 
-                    self._ip_kernel_client = BlockingKernelClient()
-                    self._ip_kernel_client.load_connection_info(self._ip_connect_info)
+                    # self._ip_kernel_client = BlockingKernelClient()
+                    # self._ip_kernel_client.load_connection_info(self._ip_connect_info)
+                    self._ip_kernel_client = self._ip_kernel_app.blocking_client()
                     logger.info(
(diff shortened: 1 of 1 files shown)


dmgav:
@kivel Thank you for the suggested improvements. When I was working on the implementation of the IP kernel feature, I started with the 'textbook' approach of creating the IP kernel app and then the client, but after a long period of testing on different systems and in different conditions I reversed the order so that the client is created and started first and then the kernel is created after some pause so that the client is ready to receive all the messages that are sent by the client. I don't remember the details of the issues I observed, but I am not comfortable to roll back the changes without significant testing. The unit tests do not reproduce all startup failure modes, so they are not sufficient.

I understand that the main problem you are trying to solve is to provide ability to specify fix ports for IP kernel so that you could use the port numbers in firewall configuration. I did not have much success with fixed port numbers, since the IP kernel was not very good at closing the ports and the operation of closing and then opening the environment very often failed. So I did not consider this mode of operation useful and did not implement the respective configuration parameters. The parameters that allow to specify fixed ports can be easily added if necessary with very minor changes to the existing startup code.

Another question for discussion is whether network configuration parameters, such as 0MQ port numbers for IP kernel, should be specified in `ipython_kernel_config.py` or should be part of Queue Server configuration. In most deployments, the IPython profile directory containing `ipython_kernel_config.py` is not part of Queue Server deployment and the config file is editable by random people, sometimes including users, while the Queue Server configuration file is deployed to a safe location and can not be easily modified.

kivel:
My main concern is that the `ipython_kernel_config.py` is consumed regardless of what you do in the code. In the simplest case, a connection file is defined, which will put the ports used at first startup into this file. Every subsequent startup will try to bind to the same ports again. Since the ports are generated randomly, this will cause hard to diagnose errors in the startup. Any other use case, like firewalls and the likes, are additional benefits, currently unachievable.
Also, the adherence to the connection file is a strict requirement in case you want to establish a persistent connection with jupyter lab to the kernel. Currently, as lined out in the documentation, one has to obtain the connection parameters and reconnect every time the environment is close/opened.

I understand that the access to the kernel configuration is troublesome, but it's a feature of ipython. I can't say that I tested this on many systems over a prolonged period. On those system I tested it, I neither had issues with unclosed ports nor any other issues.

Given the importance of the queueserver in the bluesky ecosystem. Is there a suggestion from your side how the issue can be addressed?
Imho, providing ports trough a separate config file, is not an option as it will require changing it in two places, and also break the standard config via the `ipython_kernel_config.py`.
Would an additional argument like `--unsafe-config` be reasonable? Or perhaps `--use-ipython-kernel=CONFIG`, to allow current and "textbook" behaviour?

dmgav:
What I am proposing as a simple solution is to add 5 more parameters (both command line and yaml QS config file or just yaml config file parameters) that allow to specify fixed ports. Those ports will be part of Queue Server configuration. I understand that in this case the respective parameters from `ipython_kernel_config.py` are ignored. I understand that the connection file is saved by the kernel in any case and it contains the port numbers that are actually used. So just starting Jupyter console on the same host under the same account will automatically work.

dmgav:
Closing this PR in favor of https://github.com/bluesky/bluesky-queueserver/pull/317

Issue #327, opened 2025-06-25
Children of ohphyd-async devices are not resolved properly in plans


canismarko:
Consider the following plan:

```json
{
    "name": "mv", 
    "args": ["sim_motor_2.user_setpoint", -20]
}
```

If `sim_motor_2` is a device created during the startup script with the following:

```python
from ophyd_async.epics.motor import Motor

with init_devices():
    sim_motor_2 = Motor("25idc:simMotor:m2", name="sim_motor_2")
```

then I would expect this description to result in:
```python
RE(mv(sim_motor_2.user_setpoint, -20))
```

However I get the following exception: `AttributeError: 'str' object has no attribute 'parent'`

In the existing plans and devices YAML file I see:

```yaml
  sim_motor_2:
    classname: Motor
    is_flyable: true
    is_movable: true
    is_readable: true
    module: ophyd_async.epics.motor

```

However, there are no components. An equivalent device using ophyd produces the following:

```yaml
  sim_motor_2:
    classname: EpicsMotor
    components:
      acceleration:
        classname: EpicsSignal
        is_flyable: false
        is_movable: true
        is_readable: true
        module: ophyd.signal
      ...
```

Best I can tell, this causes [`_is_object_name_in_list(v)`](https://github.com/bluesky/bluesky-queueserver/blob/ccbfd68a3fe6ba1bea5015805a6e673fc186b9f6/bluesky_queueserver/manager/profile_ops.py#L904) to fail and the device does not get resolved from it's name `"sim_motor_2.user_setpoint"` to the actual device to be moved.

Looking ahead, the following line should call `_get_nspace_object`. I'm not sure if this will work either, since this function [relies](https://github.com/bluesky/bluesky-queueserver/blob/ccbfd68a3fe6ba1bea5015805a6e673fc186b9f6/bluesky_queueserver/manager/profile_ops.py#L821) on the device having `component_names`, which ophyd-async devices don't have.

dmgav:
I will look into this issue (and https://github.com/bluesky/bluesky-queueserver/issues/328) shortly. This should not be difficult to fix.

Issue #328, opened 2025-06-25
Children of ophyd-async devices that are not readable do not get included in devices


canismarko:
Consider the following device in the queueserver startup:

```python
from ophyd_async.epics.motor import Motor
from ophyd_async.core import init_devices, Device

class SimStage(Device):
    def __init__(self, prefix: str, name = ""):
        self.vertical = Motor(f"{prefix}m2")
        self.horizontal = Motor(f"{prefix}m3")
        super().__init__(name=name)

with init_devices():
    sim_stage = SimStage("25idc:simMotor:", name="asim_stage")
```

The following should be valid to put on the queue since sim_stage.vertical satisfies `bluesky.protocols.Movable`.

```python
{
    "name": "mv", 
    "args": ["sim_stage.vertical", -20]
}
```

However, `sim_stage` does not appear in the resulting existing devices and plans file, because the parent device `sim_stage`does not satisfying either of `bluesky.protocols.Readable` or `bluesky.protocols.Flyable`: https://github.com/bluesky/bluesky-queueserver/blob/ccbfd68a3fe6ba1bea5015805a6e673fc186b9f6/bluesky_queueserver/manager/profile_ops.py#L676

If I add `protocols.HasName` to the list of checks in `is_device()`, then the sim_stage shows up in the list of devices:

```yaml
  sim_stage:
    classname: SimStage
    is_flyable: false
    is_movable: false
    is_readable: false
    module: __main__
```

Then together with #327, the original plan should work.

In practice, most of my parent devices are actually Readable, so maybe this isn't worth fixing, but maybe this should at least be documented somewhere? Or a warning issued if registering a non-readable device?

PR #329, opened 2025-07-14
Support for components (children) of ophyd-async devices

dmgav:
The PR resolves the issues https://github.com/bluesky/bluesky-queueserver/issues/327 and https://github.com/bluesky/bluesky-queueserver/issues/328.

Now the Queue Server can properly handle children of Ophyd Async devices and include then in the lists of existing and allowed devices. Also, names of the children of the Ophyd Async devices can now be passed as plan parameters and Queue Server properly converts the names (strings) to the device objects from the namespace.

### Added

- Support for children of Ophyd Async devices.

## How Has This Been Tested?

Test cases with Ophyd Async devices are added to existing tests.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -9,13 +9,11 @@ jobs:
     strategy:
       matrix:
         use-ipykernel: [false, true]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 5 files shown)


canismarko:
Tested this with both the situations in #327 and #328. Both are now fixed with the changes in this PR. Thanks @dmgav!

PR #330, opened 2025-07-23
Add a command to trim the history instead of clearing it completely

flowln:
This is intended as a less nuclear alternative to `history_clear`, allowing you to discard some older history while maintaining the newer, potentially more useful, records.

### Added
Add a command to trim the history instead of clearing it completely (`history_trim`).


dmgav:
The API that support editing the queue are naturally more extensive. Operations on history are very limited. I think it is better to extend `queue_clear` API in this case. Internally we can use separate methods to execute the operation, for example keeping `trim_history` method in `plan_queue_ops.py` is fine.

PR #343, opened 2026-01-23
Add ZMQ API message for requesting the current state of the RunEngine Metadata

jwlodek:
## Description
The Run Engine metadata is often used to store important state about the currently running experiment. For example, we use it to store the current cycle and proposal number as well as the current transient scan ID. Beamlines have requested having the ability to see this information from an interface to QueueServer (specifically queue-monitor). The new API message allows for asking for the current state of the metadata provided the environment is open.

## Motivation and Context

Beamlines wish to see the value of certain metadata keys in the UI the use for queueserver (eg. transient scan id, proposal number etc). Subsequent PRs can add the ability to use this message in the python and http apis, and eventually queuemonitor and other queueserver clients.

### Added
* New handlers for processing the `re_metadata` message to the RE manager and RE worker classes.
* New `qserver re metadata` CLI command.
* Unit test for the handlers and API.

### Changed
* Unit test for the CLI to include the new command.
* Example profile to add a sample key/value to the RunEngine metadata to start.

## How Has This Been Tested?

Manually tested:

```
(venv) $ qserver re metadata
Arguments: ['re', 'metadata']
13:26:55 - MESSAGE: 
{'success': False,
 'msg': 'Environment does not exist. Cannot retrieve Run Engine metadata.',
 're_metadata': {}}
(venv) $ qserver environment open
Arguments: ['environment', 'open']
13:27:04 - MESSAGE: 
{'success': True, 'msg': ''}
(venv) $ qserver re metadata
Arguments: ['re', 'metadata']
13:27:10 - MESSAGE: 
{'success': True,
 'msg': '',
 're_metadata': {'metadata_key': 'metadata_value',
                 'versions': {'ophyd': '1.11.0',
                              'ophyd_async': '0.14.2',
                              'bluesky': '1.14.6'}}}
```

And added in unit tests.


bluesky_queueserver/manager/tests/test_zmq_api_base.py
```diff
@@ -5971,6 +5971,69 @@ def test_zmq_api_re_runs_1(re_manager_pc_copy, tmp_path, test_with_manager_resta
...
+    assert wait_for_condition(time=5, condition=condition_environment_created)
+
+    # Get the initial run_engine metadata
+    resp, _ = zmq_request("re_metadata")
+    assert resp["success"] is True, f"{resp =}"
+
+    def check_initial_metadata(re_md):
+        """Function that checks if initial metadata matches what we expect"""
+
+        assert "metadata_key" in re_md, "Expected key not found in RE metadata"
```
dmgav: The error messages in the test are not helpful, since it is clear in which line the test failed. It is better to just print `re_md`, e.g. `pprint.pformat(re_md)`.

dmgav:
Another point worth investigation is how the server behaves if RE metadata dictionary is not serializable. Ideally the request should fail with meaningful error message.

jwlodek:
I am not sure if there's a better way to check for serializability other than just trying to serialize?

dmgav:
This is what I did for the return value of an arbitrary function, which can return arbitrary result: https://github.com/bluesky/bluesky-queueserver/blob/934008cb70b12c2329e50807f02d3347cb3f5030/bluesky_queueserver/manager/worker.py#L551

I believe the message is pickled when it is passed between processes, so the results of JSON serialization are not used.

bluesky_queueserver/manager/manager.py
```diff
@@ -3351,6 +3364,41 @@ async def _re_runs_handler(self, request):
...
+                re_metadata, msg = await self._worker_request_runengine_metadata()
+
+                if re_metadata is None:
+                    success, re_metadata = False, {}
+                else:
+                    # Make sure the metadata is serializable given the encoding
+                    if self._zmq_encoding == ZMQEncoding.JSON:
+                        json.dumps(re_metadata)
+                    else:
+                        msgpack.packb(re_metadata)
```
dmgav: Probably we should catch the exception and generate a custom error message, which includes the error message passed with the exception. The unit test should probably check for the custom part of the message.

jwlodek:
@dmgav I don't think the test failures are related to the changes here. Is this OK to merge or is there something else you wanted me to address first?

bluesky_queueserver/manager/tests/test_zmq_api_base.py
```diff
@@ -5971,6 +5971,101 @@ def test_zmq_api_re_runs_1(re_manager_pc_copy, tmp_path, test_with_manager_resta
     assert wait_for_condition(time=5, condition=condition_environment_closed)
 
 
+# fmt: off
+@pytest.mark.parametrize("test_with_manager_restart", [False, True])
+# fmt: on
+def test_zmq_api_re_metadata_1(re_manager_pc_copy, tmp_path, test_with_manager_restart):  # noqa: F811
```
dmgav: The test does not seem to be using `test_with_manager_restart` parameter.

bluesky_queueserver/manager/tests/test_zmq_api_base.py
```diff
@@ -5971,6 +5971,101 @@ def test_zmq_api_re_runs_1(re_manager_pc_copy, tmp_path, test_with_manager_resta
...
+    assert wait_for_condition(time=5, condition=condition_environment_closed)
+
+
+_add_datetime_to_md = """
+from datetime import datetime
+RE.md['date'] = datetime.now()
+"""
+
+
+def test_zmq_api_re_metadata_non_serializable_md(re_manager_pc_copy, tmp_path):  # noqa: F811
```
dmgav:
```suggestion
def test_zmq_api_re_metadata_2_non_serializable_md(re_manager_pc_copy, tmp_path):  # noqa: F811
```

bluesky_queueserver/manager/tests/test_zmq_api_base.py
```diff
@@ -5971,6 +5971,101 @@ def test_zmq_api_re_runs_1(re_manager_pc_copy, tmp_path, test_with_manager_resta
...
+    resp, _ = zmq_request("environment_open")
+    assert resp["success"] is True, f"{resp =}"
+    assert wait_for_condition(time=5, condition=condition_environment_created)
+
+    resp, _ = zmq_request("re_metadata")
+    assert resp["success"] is False
+
+    # Check that the error message is correct depending on encoding
+    if encoding == "json":
+        assert resp["msg"].startswith("Failed to serialize RE metadata with JSON:")
```
dmgav:
```suggestion
        assert resp["msg"].startswith("Failed to serialize RE metadata with JSON:"), resp
```

bluesky_queueserver/manager/tests/test_zmq_api_base.py
```diff
@@ -5971,6 +5971,101 @@ def test_zmq_api_re_runs_1(re_manager_pc_copy, tmp_path, test_with_manager_resta
...
+    elif encoding == "msgpack":
+        assert resp["msg"].startswith("Failed to serialize RE metadata with MSGPACK:")
```
dmgav:
```suggestion
        assert resp["msg"].startswith("Failed to serialize RE metadata with MSGPACK:"), resp
```

dmgav:
We need to handle the situations when there is no RE in the environment, RE does not have `md` attribute or `RE.md` can not be converted to a dictionary. Probably the request should fail with useful message if there is no RE in the environment and return `{}` in the other cases. For testing, `RE` can be removed from the existing environment or replaced by executing a script, e.g.:
```
    resp2, _ = zmq_request("script_upload", params={"script": "delete RE"})
    assert resp2["success"] is True
    result = wait_for_task_result(time=10, task_uid=resp2["task_uid"])
    assert result["success"] is True, pprint.pformat(result)
```
To check that RE was removed successfully it may be useful to check `re_state` field of the status (it should be None).

dmgav:
Since we are adding a new API, we should also update the API documentation, may be put the API description after the description of `re-runs` API : https://blueskyproject.io/bluesky-queueserver/re_manager_api.html#re-runs

docs/source/re_manager_api.rst
```diff
@@ -1676,7 +1680,7 @@ Description   Request the list of active runs generated by the currently execute
               included in the returned parameters
 ------------  -----------------------------------------------------------------------------------------
 Parameters    **option**: *'active'*, *'open'* or *'closed'* (optional)
-                  select between full list of 'active' (default) runs, the list of 'open' or 'closed'
+                  Select between full list of 'active' (default) runs, the list of 'open' or 'closed'
```
dmgav:
```suggestion
                  select between full list of 'active' (default) runs, the list of 'open' or 'closed'
```

docs/source/re_manager_api.rst
```diff
@@ -1686,7 +1690,7 @@ Returns       **success**: *boolean*
                   error message in case of failure, empty string ('') otherwise.
 
               **run_list**: *list(dict)*
-                  the requested list of runs, list items are dictionaries with keys 'uid' (str),
+                  The requested list of runs, list items are dictionaries with keys 'uid' (str),
```
dmgav:
```suggestion
                  the requested list of runs, list items are dictionaries with keys 'uid' (str),
```

dmgav (approved):

PR #344, opened 2026-01-23
Since the tests take a while, cancel any-in-progress tests when a new…

jwlodek:
… commit is pushed. Only workflows triggered on the same PR will be cancelled (i.e. workflows triggered by a second PR will be unaffected).


jwlodek:
This seems to have worked - you can see in the first commit's CI report that all the unit test jobs got cancelled.

dmgav (approved):

PR #345, opened 2026-03-02
Add additional configuration option to allow for filtering which RE metadata keys can be read via the ZMQ API.

jwlodek:
- **Add utility function for filtering dictionary based on list of permitted keys**
- **Add additional config option to list permitted RE metadata keys.**
- **Add unit tests for re metadata filtering**
- **Apply pre-commit**

Tested via included unit tests, and via the CLI:

```
(venv) jwlodek@alma10:~/Workspace/bluesky-queueserver$ qserver environment open
Arguments: ['environment', 'open']
10:08:29 - MESSAGE: 
{'success': True, 'msg': ''}
(venv) jwlodek@alma10:~/Workspace/bluesky-queueserver$ qserver re metadata
Arguments: ['re', 'metadata']
10:08:32 - MESSAGE: 
{'success': True, 'msg': '', 're_metadata': {'versions': {'bluesky': '1.14.6'}}}
(venv) jwlodek@alma10:~/Workspace/bluesky-queueserver$ qserver environment open
Arguments: ['environment', 'open']
10:09:56 - MESSAGE: 
{'success': True, 'msg': ''}
(venv) jwlodek@alma10:~/Workspace/bluesky-queueserver$ qserver re metadata
Arguments: ['re', 'metadata']
10:09:59 - MESSAGE: 
{'success': True,
 'msg': '',
 're_metadata': {'versions': {'ophyd': '1.11.0', 'bluesky': '1.14.6'}}}
```


bluesky_queueserver/manager/config.py
```diff
@@ -487,6 +488,13 @@ def _check_port_value(port, port_name):
             self._settings["ipython_control_port"], "control"
         )
 
+        self._settings["permitted_re_metadata_keys"] = self._get_param(
+            value_default=args.permitted_re_metadata_keys,
+            value_config=self._get_value_from_config("permitted_re_metadata_keys"),
+            value_ev=os.environ.get("QSERVER_PERMITTED_RE_METADATA_KEYS", "/").split(","),
```
jwlodek: Should we split on comma here, or `:` to make it more like most env vars that have lists of paths?

dmgav: In my opinion it is better to use `:`

bluesky_queueserver/manager/utils.py
```diff
@@ -30,3 +30,47 @@ def generate_random_port(ip=None):
     sock.close()
 
     return port
+
+
+def filter_dict_by_permitted_keys(d, permitted_keys):
+    """
+    Filter the dictionary `d` by the list of `permitted_keys`. The keys in the dictionary that are not in the list
+    of permitted keys will be removed. The keys in the list of permitted keys can be specified as absolute paths
+    (e.g. "/key1/key2") or as prefixes (e.g. "/key1/"). In the latter case, all keys that start with the prefix
```
dmgav: Is there a particular reason why we can't use `*` to specify a prefix? Using `/` may be counter-intuitive if we state that the keys are represented in the format used for Linux paths. The trailing `/` should probably be ignored if someone accidentally adds it.

jwlodek: The behavior is basically that the trailing path is ignored. I'll update the docstring to remove the indication of a prefix.

jwlodek:
I've removed the confusing wording and just remove trailing slashes now. I also updated the docs.

PR #356, opened 2026-08-04
Remove several deprecated CLI flags, and any associated references

jwlodek:
The CLI flags in question have been deprecated for at least several releases, and thus should be safe to remove

## Description

Removes deprecated CLI flags for keep-re and the zmq-info address port

## Motivation and Context
Cleaning up deprecated features reduces the size of the codebase and removes unnecessary conditionals and warning messages

### Removed
Deprecated CLI flags for setting the ZMQ info address and for keeping the RE.

## How Has This Been Tested?
Unit/Integration tests


jwlodek:
For context, `keep_re` has been deprecated for the last three releases and over a year, while the others have been deprecated for over four years.

dmgav (approved):

PR #357, opened 2026-08-04
Support streaming RE waiting hook watcher updates over the info ZMQ topic

### LINKED ISSUE #353: Allow for streaming RE watcher updates to remote clients via zmq/websocket
The RE has a progress `waiting_hook` API that allows for clients to display progress in a given run/scan.

A new `waiting_hook` compatible class should be added that the worker can optionally use to serialize these updates and stream them alongside the status and info streams to clients. Clients can then display progress bars and the like, just like the CLI progress bar manager.

jwlodek:
Resolves #353. Adds an additional keyed stream to the info zmq topic for streaming RE waiting hook progress updates. Adds additional CLI/config options to enable the feature + a waiting hook compatible class that posts the updates from the RE to zmq. Clients can then use these updates to construct progress bars and the like.

## Motivation and Context
Currently, a common piece of feedback from BL staff that use queueserver is that there isn't enough feedback on the state of a scan as it is taking place. Being able to re-create something similar to the bluesky `ProgressBarManager` by using streamed RE updates should ameliorate this issue.

### Added
- Implementation for streaming of device progress data.
- Additional CLI parameter of `start-re-manager`: `--zmq-stream-device-progress` and `network/zmq_stream_device_progress` config file equivalent.

### Changed
- Separate CLI parameter for controlling Info message publishing: `--zmq_publish_info` (`network/zmq_publish_info` config file parameter).

## How Has This Been Tested?

Added unit tests, and explicit local integration test

### DIFF
--- a/src/bluesky_queueserver/manager/output_streaming.py
+++ b/src/bluesky_queueserver/manager/output_streaming.py
@@ -115,29 +115,8 @@ def push_info_to_msg_queue(*, key, msg, msg_queue):
     msg_queue.put(msg)
 
 
-def push_progress_to_msg_queue(*, msg, msg_queue):
-    """
-    Format a progress message and put it into the message queue. The message is published
-    to the ``progress`` channel on the ``QS_Progress`` 0MQ topic.
-
-    Parameters
-    ----------
-    msg : dict
-        The progress payload dictionary.
-    msg_queue : multiprocessing.Queue
-        Reference to the queue used for collecting messages.
-
-    Returns
-    -------
-    None
-    """
-    msg = {"channel": "progress", "time": ttime.time(), "msg": msg}
-    msg_queue.put(msg)
-
-
 _default_zmq_console_topic = "QS_Console"
 _default_zmq_info_topic = "QS_Info"
-_default_zmq_progress_topic = "QS_Progress"
 
 
 class PublishZMQStreamOutput:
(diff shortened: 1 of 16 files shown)


dmgav:
In my opinion, the progress data needs to be dumped into the existing QS_info topic. Adding a separate topic does not make much sense, since the data will most likely need to be merged at the HTTP Server and streamed on the same web socket. Also no changes are required to HTTP server and the API package if the same topic is reused.

src/bluesky_queueserver/manager/output_streaming.py
```diff
@@ -115,8 +115,29 @@ def push_info_to_msg_queue(*, key, msg, msg_queue):
     msg_queue.put(msg)
 
 
+def push_progress_to_msg_queue(*, msg, msg_queue):
```
dmgav, line 118: If the information is published to the QS_info topic, use `push_info_to_msg_queue`. The `key` could be `motor_watcher` or similar. The key is used to distinguish between messages.

src/bluesky_queueserver/manager/start_manager.py
```diff
@@ -624,6 +623,25 @@ def formatter(prog):
...
+        "--zmq-publish-info",
+        dest="zmq_publish_info",
+        type=str,
+        choices=["ON", "OFF"],
+        default="ON",
+        help="Enable (ON) or disable (OFF) publishing of info/status updates to 0MQ (default: %(default)s).",
+    )
+
+    group_console_output.add_argument(
+        "--zmq-publish-progress",
```
dmgav, line 636: The parameter probably should be `--zmq-stream-motor-watcher` or similar to distinguish between the parameters enable/disable all streaming for the whole topic.

PR #361, opened 2026-08-10
Add opt-in Sardana mode to RE Manager

udaisingh93:
### Summary

Adds an opt-in mode in which RE Manager drives a **Sardana** Pool/MacroServer/Door
over Tango instead of a local bluesky `RunEngine`. Sardana macros are exposed
through the existing queue/plan API, so `qserver` clients and web UIs work against
a Sardana beamline without changes.

The mode is **off by default and fully gated** — with `--sardana-enable OFF`
(the default) no Tango import, connection, or code path is reached, and existing
RE behavior is byte-for-byte unchanged.

### Motivation

Facilities whose hardware control stack is Sardana/Tango rather than ophyd
currently cannot use queueserver at all. This lets them reuse the queue, history,
console streaming, and permissions machinery while execution happens on a Door.

### Enabling

```bash
start-re-manager --sardana-enable ON --tango_url localhost:10000
```

`--sardana-enable ON` is the only switch that turns the mode on. The Tango database
address resolves from `--tango_url` → `run_engine/tango_url` config → `TANGO_HOST`
env var; if none resolves, startup fails with a clear `ConfigError`. Setting
`--tango_url`/`TANGO_HOST` alone never enables the mode.

### What changed

**`worker.py`** — the bulk of the integration:
- `existing_plans` are built live from the MacroServer: `MacroList` plus a single
  bulk `GetMacroInfo` call (~7 ms for 300+ macros) converted into qserver plan
  descriptions, with a bare-stub fallback if `GetMacroInfo` fails.
- `existing_devices` come from Pool list attributes (`Motors`, `ExpChannels`,
  `IORegisters`). Measurement groups are collected for macro forms but kept out of
  `existing_devices` and surfaced via `config_get` instead.
- Element-typed macro parameters get `annotation.devices[<group>]` so UIs can render
  dropdowns; `ParamRepeat` groups are handled so alternating device/number pairs
  (e.g. `mv`) don't fail strict validation.
- Execution: `dev.runmacro(args)`, with console output streamed by subscribing to the
  Door's `output` attribute (`CHANGE_EVENT`).
- `re_state` maps Door `DevState` → qserver states: `RUNNING`→`running`,
  `STANDBY`→`paused`, `ALARM`/`FAULT`→`panicked`, otherwise `idle`.
- Pause/resume use `pausemacro()`/`resumemacro()`. Because `PauseMacro` is
  cooperative and macros that never poll a pause point can't reach `STANDBY`,
  `abort`/`stop`/`halt` are issued directly against a *running* Door
  (`abortmacro()`/`stopmacro()`) rather than requiring a paused state first.

**`manager.py`** — Sardana gating, measurement-group discovery, and interrupt routing.
The Tango connection is deliberately **not** made in `__init__`: `RunEngineManager`
subclasses `multiprocessing.Process`, and Tango objects aren't picklable, so eager
connection breaks `Process.start()` under the `spawn` start method (Windows). It is
deferred to `_connect_sardana()`, called from `run()` after the process starts.

**`config.py` / `start_manager.py`** — `--sardana-enable` and `--tango_url` flags,
settings resolution, and the config keys. `tango_url` is only propagated into the
worker/manager config when the mode is enabled.

### Testing

`src/bluesky_queueserver/manager/tests/test_sardana_mode.py` — 24 unit tests covering
settings resolution, manager/worker gating (including that nothing touches Tango when
disabled), Door state translation, macro→plan schema conversion, and non-fatal handling
of an unreachable Tango DB or missing Pool. All pass on Linux alongside the existing
suite. Also verified end-to-end against a live Sardana Pool/MacroServer/Door.


whs92:
Could you add changes to the docs with instructions on how to configure this and an api description?

I think this is a great idea.

udaisingh93:
Hi, thanks for suggestions, I try to add docs after our meeting

tacaswell:
I am very skeptical of threading the sardana logic through the existing code in this way.  I think it would be better to have a sardana worker implementation sit along side the RE worker implementation.

Issue #347, opened 2026-04-23
ImportError when startup script imports bluesky


ddkohler:
When I try to build a list of devices and scans with a script, and that script imports from bluesky, e.g.

/startup.py
```python
from bluesky.preprocessors import baseline_decorator
``` 

Running `qserver-list-plans-devices --startup-script startup.py` fails, trying to import numpy again within a multiprocessing process:

```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.14/site-packages/bluesky_queueserver/manager/gen_lists.py", line 218, in run
    existing_plans, existing_devices, _, _ = existing_plans_and_devices_from_nspace(
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        nspace=nspace, ignore_invalid_plans=ignore_invalid_plans, max_depth=device_max_depth
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.14/site-packages/bluesky_queueserver/manager/profile_ops.py", line 3416, in existing_plans_and_devices_from_nspace
    devices_in_nspace = devices_from_nspace(nspace)
  File "/usr/local/lib/python3.14/site-packages/bluesky_queueserver/manager/profile_ops.py", line 765, in devices_from_nspace
    if is_device(obj):
       ~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.14/site-packages/bluesky_queueserver/manager/profile_ops.py", line 680, in is_device
    from bluesky.protocols import Flyable, Readable
  File "/usr/local/lib/python3.14/site-packages/bluesky/__init__.py", line 3, in <module>
    from .preprocessors import SupplementalData
  File "/usr/local/lib/python3.14/site-packages/bluesky/preprocessors.py", line 6, in <module>
    from bluesky.protocols import Locatable
  File "/usr/local/lib/python3.14/site-packages/bluesky/protocols.py", line 5, in <module>
    from event_model.documents import Datum, StreamDatum, StreamResource
  File "/usr/local/lib/python3.14/site-packages/event_model/__init__.py", line 33, in <module>
    import numpy
  File "/usr/local/lib/python3.14/site-packages/numpy/__init__.py", line 112, in <module>
    from numpy.__config__ import show_config
[...]

PR #368, opened 2026-09-01
Prevent standard packages from being recognized as local imports

dmgav:
## Description

If a virtual environment (`venv`, `.pixi`) resides inside the startup script or startup directory, and local imports are enabled, standard packages imported by the startup code (e.g. `numpy`) may be incorrectly removed from `sys.modules` during cleanup. Any subsequent attempt to re-import such packages fails in Python 3.13 due to restrictions on reloading C extensions. This PR fixes the issue by skipping modules whose path contains site-packages during the local module cleanup step.

In general it is recommended that in production environments the startup code is placed in dedicated directory instead of the root `/` directory of the container. This may help avoid such issues.

## Motivation and Context

Addresses the issue https://github.com/bluesky/bluesky-queueserver/issues/347

### Fixed

- Prevent standard packages from being recognized as local imports.


thopkins32 (approved):
LGTM

Issue #300, opened 2024-03-27
RunEngineResult is not JSON serializable when RunEngine flag call_returns_result=True.


burkeds:
I am running the queue server in a docker container and following along with the tutorial for the queue server. My startup code is simply the startup code in the ophyd-async tutorial (https://blueskyproject.io/ophyd-async/main/user/tutorials/using-existing-devices.html).

I am controlling the queue server via qserver bash commands. I open the environment, clear the queue and add an item. 

`qserver queue add plan '{"name": "count", "args": [["det1", "det2"]], "kwargs": {"num": 10, "delay": 1}}'`
`qserver queue start`

After the 10 counts a TypeError is raised.

`[I 2024-03-27 10:34:37,551 bluesky_queueserver.manager.manager] Starting the plan:
2024-03-27T10:34:37.551885768Z {'name': 'count',
2024-03-27T10:34:37.551904800Z  'args': [['det1', 'det2']],
2024-03-27T10:34:37.551907759Z  'kwargs': {'num': 10, 'delay': 1},
2024-03-27T10:34:37.551910601Z  'user': 'qserver-cli',
2024-03-27T10:34:37.551913431Z  'user_group': 'primary',
2024-03-27T10:34:37.551916201Z  'meta': {},
2024-03-27T10:34:37.551919128Z  'item_uid': '577a0133-772d-49fc-8dfb-fd48020aec7b'}.
2024-03-27T10:34:37.552487927Z [I 2024-03-27 10:34:37,551 bluesky_queueserver.manager.worker] Starting execution of a plan ...
2024-03-27T10:34:37.552600568Z [I 2024-03-27 10:34:37,552 bluesky_queueserver.manager.worker] Starting a plan 'count'.
2024-03-27T10:34:37.740276573Z [I 2024-03-27 10:34:37,739 bluesky_queueserver.manager.plan_monitoring] New run was open: 'dd4812f0-a081-4e5a-b6fc-cc61e03d8f85'
2024-03-27T10:34:47.761402028Z [I 2024-03-27 10:34:47,757 bluesky_queueserver.manager.plan_monitoring] Run was closed: 'dd4812f0-a081-4e5a-b6fc-cc61e03d8f85'
2024-03-27T10:34:47.762916747Z [I 2024-03-27 10:34:47,762 bluesky_queueserver.manager.worker] The plan was exited. Plan state: completed
[...]

PR #369, opened 2026-09-03
Add plan return result information to history (when serializable)

thopkins32:
Assisted-by: oh-my-pi:gpt-5.6-terra

## Description

Adds functionality to capture the plan return and attempt to attach it to the queue item's result. If the return value is not JSON serializable, then we replace it with none and give an error instead.

## Motivation and Context

Related to #300, not sure if it closes it.

I would like there to be a simple way to put a plan like `bps.rd` in the queue and inspect its result without having to use `bp.count` instead. Often I just want to know the current state of a signal without the overhead of opening a run -> staging -> reading -> saving -> unstaging -> closing a run and all of the callbacks involved with that.

Capturing the real return value of the plan and storing it in the queue item solves this problem.

### Added

- Return value of plans are captured and put in the queue item, queryable from the queue history.

## How Has This Been Tested?

Added checks to existing unit tests and wrote new ones using the zmq api.

### DIFF
--- a/docs/source/re_manager_api.rst
+++ b/docs/source/re_manager_api.rst
@@ -654,6 +654,9 @@ Returns       **success**: *boolean*
 
                   - **traceback** - full traceback if the plan failed, empty string otherwise.
 
+                  If ``capture_plan_return_values`` is enabled, newly finalized history items contain the
+                  following keys:
+
                   - **return_value** - JSON-normalized terminal return value of a plan with **exit_status**
                     **'completed'**. *None* is a valid plan return and is also used for plans without a completed
                     return value.
@@ -661,7 +664,8 @@ Returns       **success**: *boolean*
                   - **return_value_error** - empty string if the return value was retained; otherwise, the
                     JSON-serialization diagnostic while the plan remains completed.
 
-                  History items created before these fields were added may not contain either key.
+                  History items created while capture is disabled or before these fields were added do not contain
+                  either key.
 
               **plan_history_uid**: *str*
                   current plan history UID.
--- a/docs/source/re_manager_api.rst
+++ b/docs/source/re_manager_api.rst
@@ -654,6 +654,15 @@ Returns       **success**: *boolean*
 
                   - **traceback** - full traceback if the plan failed, empty string otherwise.
 
+                  - **return_value** - JSON-normalized terminal return value of a plan with **exit_status**
+                    **'completed'**. *None* is a valid plan return and is also used for plans without a completed
+                    return value.
+
+                  - **return_value_error** - empty string if the return value was retained; otherwise, the
+                    JSON-serialization diagnostic while the plan remains completed.
+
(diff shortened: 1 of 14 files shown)


src/bluesky_queueserver/manager/worker.py
```diff
@@ -512,7 +526,10 @@ def _generate_new_plan(self, parameters):
 
             def get_start_plan_func(plan_func, plan_args, plan_kwargs, plan_meta):
                 def start_plan_func():
-                    return self._RE(plan_func(*plan_args, **plan_kwargs), {"all": [self._run_reg_cb]}, **plan_meta)
+                    def plan_with_return_value():
```
dmgav: Why do we need to wrap the plan function into another plan? The Run Engine is expected to return the results if the plan contains the return statement. So the original code on the line 515 should work.

thopkins32: Technically, the return value depends on whether the RE was initially configured with `call_returns_result=True` or not. In this case, you don't have to set that to `True` or `False` and you will still be able to get the result of the generator.

It's the same reason the `uid` is not taken from the `result` but rather taken from `CallbackRegisterRun`.

docs/source/re_manager_api.rst
```diff
@@ -654,6 +654,15 @@ Returns       **success**: *boolean*
 
                   - **traceback** - full traceback if the plan failed, empty string otherwise.
 
+                  - **return_value** - JSON-normalized terminal return value of a plan with **exit_status**
```
dmgav, line 660: The message could probably be reduced a little bit:
- for plans that don't return a value or not completed the message contains `return_value: None` (or no `return_value` at all to save space),
- for plans that return a value the options are `return_value: {value: <value>}` if the value was successfully serialized, otherwise `return_value: {error: <serialization error message>}`.

We can also optionally include either `return_value` or `return_value_error` depending on what data exists.

src/bluesky_queueserver/manager/worker.py
dmgav: The `CallbackRegisterRun` is used mostly to get UIDs for real-time monitoring. I don't think we should bypass `call_returns_result` parameter. Plans can potentially return large arrays and other data that should not be put in the history. Clients should use Ophyd-as-service to reading the state of devices (essentially PV values). I would advise agains direct use of bps.rd for reading PVs, unless those reads must be queued to support the workflow.

thopkins32: There are other return values that may be of interest that are not available as PVs or ophyd signals that also don't require opening a full Bluesky run. Queued reads are not the only desired workflow here, although that is something I would like. I think bypassing `call_returns_result` is good practice. Numpy arrays are not JSON serializable, so they will not be included in the history already. We can add limits to arbitrary record sizes, but that seems premature.

thopkins32:
@dmgav should be good to review again. The CI failures seem to be existing.

PR #373, opened 2026-09-18
Include enums among information passed to clients about nspace

jwlodek:
Passes enum names, values to clients along with plans and devices

## Description
Checks namespace items to see if they are enums, and serializes and streams them to clients.

## Motivation and Context
Enum values are often used in UIs to denote a list of valid choices. It would be useful to know the possible states for enum parameters to plans from a client.

## How Has This Been Tested?
Unit tests for now. Going to have an integration test shortly.


dmgav:
Considering that most of the Queue Server clients are web applications, it doesn't seem right to for them to delegate processing of python headers to the client. There must be another way to solve this problem without adding extra API and extra parameter to status. Clients do not need to know about all the enums defined in the profile.

jwlodek:
> Clients do not need to know about all the enums defined in the profile

What would you propose? Ultimately, if a plan takes an enum argument, the client must know the possible choices available to display a reasonable dropdown button to a user. I don't see another way of getting all of this information to the client other than encoding the possible values. Perhaps they can be encoded in the plan headers, but I think you'd end up with lots of duplicated information, if plans all take the same enum type.

